PR #11, opened 2023-01-24
enhancement: BaseServerAddon.get_client_source_info now returns stric…

martastain:
`get_client_source_info` is now expected to return one of the three stict models for source info, which are defined in `ayon_server.addons.models`

If the returned type is different, it won't crash the request, but the problem is logged. 

This change breaks the current implementation of the site sync addon, which already reimplements get_client_source_info.

Note that models are not yet well-typed and documented. @iLLiCiTiT feel free to add descriptions or modify the models.

### DIFF
--- a/api/addons/__init__.py
+++ b/api/addons/__init__.py
@@ -128,7 +128,9 @@ async def list_addons(
 
                 elif not all([isinstance(x, SourceInfo) for x in source_info]):
                     logging.error(f"Invalid source info for {addon.name} {version}")
-                    source_info = []
+                    source_info = [
+                        x for x in source_info if isinstance(x, SourceInfo)
+                    ]
                 vinf["client_source_info"] = source_info
 
                 vinf["services"] = addon.services or None
--- a/api/addons/__init__.py
+++ b/api/addons/__init__.py
@@ -6,6 +6,7 @@ from nxtools import logging
 
 from addons import project_settings, site_settings, studio_settings
 from ayon_server.addons import AddonLibrary
+from ayon_server.addons.models import SourceInfo
 from ayon_server.api.dependencies import dep_current_user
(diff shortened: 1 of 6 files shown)


api/addons/__init__.py
```diff
@@ -125,9 +121,16 @@ async def list_addons(
...
-                )
+
+                source_info = await addon.get_client_source_info(base_url=base_url)
+                if source_info is None:
+                    pass
+
+                elif not all([isinstance(x, SourceInfo) for x in source_info]):
+                    logging.error(f"Invalid source info for {addon.name} {version}")
+                else:
+                    vinf["client_source_info"] = source_info
```
iLLiCiTiT, line 132: There would be missing "client_source_info" key.
```suggestion
                    logging.error(f"Invalid source info for {addon.name} {version}")
                    source_info = []

                vinf["client_source_info"] = source_info
```

ayon_server/addons/models.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class PathDefinition(OPModel):
+    plaform: str
+    path: str
+
+
+class FilesystemSourceInfo(ClientSourceInfo):
+    type: Literal["filesystem"] = Field("filesystem")
+    paths: list[PathDefinition] = Field(default_factory=list)
```
iLLiCiTiT, line 17: I realized this morning that this actually does't make sence to have it as list.
```suggestion
class PathDefinition(OPModel):
    windows: str
    linux: str
    darwin: str


class FilesystemSourceInfo(ClientSourceInfo):
    type: Literal["filesystem"] = Field("filesystem")
    path: PathDefinition = Field(default_factory=PathDefinition)
```

iLLiCiTiT (approved):

PR #17, opened 2023-02-27
Use universal bash definition

iLLiCiTiT:
## Description
Use recommended `/usr/bin/env bash` instead of `/bin/bash` in shell scripts.

## Additional information
This fixes issue on windows launching the scripts in docker.


martastain:
Good point. Thanks!

PR #18, opened 2023-04-11
Settings: Function 'normalize_name' has more options

iLLiCiTiT:
## Description
Modified `normalize_name` to match most of usecases in settings and give more options to do when needed.


martastain:
Replacing allowed chars would not work with the default. As the whitelist is escaped, the regex would become `[^a\-zA\-Z0\-9\-_\ ]` and `normalize_name("test")` would return an empty string. Just use `allower_chars_regex = f"[^{allowed_chars}]"`

martastain (approved):
passes my tests

PR #19, opened 2023-04-14
Refactor: Annotated dependencies

martastain:
This pull request changes the way how REST dependencies are handled. It utilizes new type annotations of FastAPI both for dependencies and return types, which is easier to read and makes the code more maintainable. Original dep_* functions are still available, but for each dependency there is a Annotated type, which may be used instead.


iLLiCiTiT (approved):
Tried few things and looks like it works...

api/tasks/tasks.py
```diff
@@ -90,19 +78,15 @@ async def create_task(
 #
 
 
-@router.patch(
-    "/projects/{project_name}/tasks/{task_id}",
-    status_code=204,
-    response_class=Response,
-)
+@router.patch("/projects/{project_name}/tasks/{task_id}")
```
Minkiu, line 81: Should status code be declared here too?

api/projects/anatomy.py
```diff
@@ -56,13 +50,13 @@ async def get_project_anatomy(
     )
 
 
-@router.post("/projects/{project_name}/anatomy", response_class=Response)
+@router.post("/projects/{project_name}/anatomy", status_code=204)
```
Minkiu, line 53: I believe this is the default status code for `EmptyResponse`

api/events/enroll.py
```diff
@@ -55,13 +54,16 @@ class EnrollResponseModel(OPModel):
 @router.post("/enroll", response_model=EnrollResponseModel)
```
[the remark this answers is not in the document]
martastain: I don't think so. 204 is not the "Primary" response. What you expect is EnrollResponseModel and 204 is treated more or less like 404 (but error code is not used since it would flood the logs). But response_model argument is redundant here.

PR #20, opened 2023-05-26
Rename subset to product

martastain:
Breaking change in the database schema and API

 - Renamed `subset` to `product`
 - Renamed `family` to `product_type` / `productType`
 - Alias `node.type` on graphql folder, task and product nodes
 - `public.product_types` table collecting all posted productTypes


ayon_server/graphql/resolvers/tasks.py
```diff
@@ -71,7 +71,7 @@ async def get_tasks(
         return TasksConnection(edges=[])
 
     project_name = root.project_name
-    fields = FieldInfo(info, ["tasks.edges.node", "subset"])
+    fields = FieldInfo(info, ["tasks.edges.node", "task"])
```
iLLiCiTiT, line 74: This was a bug before, or is bug now? (`subset` -> `task`)

martastain: It was a bug before. This is used to check, which fields are requested in the query. The bug occurred rarely in the UI, so it was there for a long time - basically it caused folder attributes weren't propagated to task if the task was requested directly using its id.

iLLiCiTiT:
Tried what I could, and all seems to work, for now.

PR #27, opened 2023-06-15
Bug: fix line ends in start.sh

antirotor:
## Description

Just fixing EOL issue in `start.sh` that was crashing docker container on start - converting from windows to linux EOL.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -21,3 +21,4 @@ exec gunicorn \
   --log-level warning \
   -b :5000 \
   ayon_server.api:app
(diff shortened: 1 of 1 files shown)


martastain:
I know this is a recurring issue (and it is why I asked you to test on windows), but how does this happen? if I understand correctly the file on your side has windows line endings, but I've created it on linux and I don't use autoclrf. does windows git change line endings of cloned files to windows? what line endings the rest of the files have?

PR #29, opened 2023-06-20
Replace Addon's Library logic with Rez

Minkiu:
Replace `addons` logic with Rez

This Pull Request, modifies the backend in order to leverage Rez to manage AYON addons (they should now include a `package.py` with the attribute [`plugin_for="ayon_server"`](https://github.com/AcademySoftwareFoundation/rez/wiki/Package-Definition-Guide#plugin_for).

A loose explantation of the process is:
1) When we spin the server, we make sure a (meta) package named `ayon_server` exists in our repo, as of now `/addons`; and it matches the current `ayon-backend` version, otherwise we create one, and request a restart, this package has the attribute [`has_plugins=True`](https://github.com/AcademySoftwareFoundation/rez/wiki/Package-Definition-Guide#has_plugins)
2) We try to find plug-ins for our `ayon_server`.
3) We, as before, attempt to initialize the add-ons defined in the bundles; we can now list all installed packages without actually initializing them, and keep a list of initialized and broken ones.
4) When installing an addon, we have a `rezbuild.py` that we fall back for simple addons; there's a bug in rez that makes it mandatory for the `package.py` to contain the `build_command` attribute set (tried to add it programmatically, but it won't install the package, due this bug).

All the rest should work as it does now, I really haven't tested the migration of settings between bundles, so that might not be working at the minute.

## Reviewing
I tired to keep it tidy and compartemantilized, so there are 4 commits:
1) Implement the RezRepo class
2) Refactor AddonLibrary to make use of the RezRepo
3) Modify all instance of the AddonLibrary in the `api` folder, to adapt to any required changes
4) Modify all instances of the AddonLibrary in the `ayon_server` folder, to adapt to any required changes, this one has some bits that have changed more than simple method renames.

## TODO
* Adapt AYON addons to rez packages (simply add a `package.py` at the root).
* Refactor the `BaseServerAddon` and move some of the attributes to the `package.py` then, read the info from there.

## Testing note
I would suggest making a copy of the `docker-compose.yml` in the `ayon-docker` backend and build a new image called `ayon-rez` that build the new `rez` dependency; so it does not interfere with your existing deploys; also map the `/addons` folder to a different one than the one you are currently using.

```yml
version: "3.7"

services:
  postgres:
    image: postgres:15
    restart: unless-stopped
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U ayon"]
      interval: 5s
      timeout: 5s
      retries: 5

    expose: [5432]

    volumes:
      - "/etc/localtime:/etc/localtime:ro"
      - "db:/var/lib/postgresql/data"

    environment:
      - "POSTGRES_USER=ayon"
      - "POSTGRES_PASSWORD=ayon"
      - "POSTGRES_DB=ayon"

  redis:
    image: redis:alpine
    restart: unless-stopped
    healthcheck:
      test: ["CMD-SHELL", "redis-cli ping | grep PONG"]
      interval: 5s
      timeout: 5s
      retries: 5

    expose: [6379]


  server:
    image: localhost/ayon-rez:latest
    restart: unless-stopped
    healthcheck:
      test: ["CMD", "curl", "-f", "http://localhost:5000/api/info"]
      interval: 10s
      timeout: 2s
      retries: 3

    depends_on:
      postgres:
        condition: service_healthy
      redis:
        condition: service_started
    # uncomment for older versions of Docker (>4.0.0) and comment out above "depends on" section.
    #depends_on:
      #- postgres
      #- redis

    expose: [5000]
    ports: ["5000:5000"]

    volumes:
      - "./rezrepo:/addons:Z"
      - "./storage:/storage:Z"

      # comment out the following line on Windows
      - "/etc/localtime:/etc/localtime:ro"

      # uncomment the following line if you need to work on the backend code
      - "/home/minkiu/Projects/ynput/ayon-backend:/backend:Z"

volumes:
  db: {}
```
You can use this PR to generate rez packages out of the OpenPype ones: https://github.com/ynput/OpenPype/pull/6169

### DIFF
--- a/ayon_server/addons/rezbuild.py
+++ b/ayon_server/addons/rezbuild.py
@@ -8,7 +8,7 @@ This expects a certain folder structure:
 
 Other folders might or might not be present, but they are essentially ignored.
 
-Will zip up the `client/ayon_<addon_name>` and place it into `private/client.zip`
+Will zip up the `client/<client module name>` and place it into `private/client.zip`
 and copy the contents of `server` into the root of the package.
 
 <package root>


Minkiu:
Marking as draft so we can have a conversation about this.

BigRoy:
Thanks for drafting a prototype! I haven't been involved much with the Ayon backend but just wanted to say that any work on this front in combination with Rez is greatly appreciated.

This PR's Rez integration is solely regarding the addons and modules in the backend, not the running of the actual DCC application and tools, right? So it's not for those that might have a huge ready-to-go Rez repository already with many applications and tools and allow them to launch those through OpenPype for certain hosts. Or am I mistaken?

antirotor:
> This PR's Rez integration is solely regarding the addons and modules in the backend, not the running of the actual DCC application and tools, right? So it's not for those that might have a huge ready-to-go Rez repository already with many applications and tools and allow them to launch those through OpenPype for certain hosts. Or am I mistaken?

For that we will need to implement alternative to applications addon that will use rez and also we will need to create rez package from ayon-launcher (that will be the new entry point as Igniter in OP is)

BigRoy:
> For that we will need to implement alternative to applications addon that will use rez and also we will need to create rez package from ayon-launcher (that will be the new entry point as Igniter in OP is)

Great. Just to confirm, so this particular PR is not intended to do **that**. Right?

Minkiu:
You are right, this is not the PR for that.

ayon_server/addons/addon.py
```diff
@@ -46,7 +46,7 @@ def __repr__(self) -> str:
     @property
     def friendly_name(self) -> str:
         """Return the friendly name of the addon."""
-        return f"{self.definition.friendly_name} {self.version}"
+        return f"{self.name.capitalize()}"
```
iLLiCiTiT, line 49: This should stay with the version.

ayon_server/addons/rezlibrary.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def items(self) -> ItemsView[str, BaseServerAddon]:
+        return self.initialized_addons.items()
+
+
+
+
+
+
+
```
iLLiCiTiT, line 219:
```suggestion
```

iLLiCiTiT:
Is with this PR possible to manually add/remove addon to/from server?

Minkiu:
No, this PR doesn't implement the adding/removing side of thing, will wait till the bundle changes settle to rebase and piggyback on current implementation to install /remove and hook there Rez

Minkiu:
I am finally satisfied with the commit history, and managed to catch some bugs, it is now ready to be reviewed, will update the PR to reflect changes.

api/system/info.py
```diff
@@ -88,21 +88,15 @@ async def get_sso_options(request: Request) -> list[SSOOption]:
...
 
-    for _name, definition in library.data.items():
-        try:
-            vers = active_versions.get(definition.name, {})
-        except ValueError:
-            continue
-        production_version = vers.get("production", None)
-        if not production_version:
-            continue
+    library = AddonLibrary().get_instance()
```
iLLiCiTiT, line 92:
```suggestion
    library = AddonLibrary.get_instance()
```

ayon_server/addons/rezbuild.py
```diff
@@ -0,0 +1,112 @@
+""" AYON Rez Builder
+
+This expects a certain folder structure:
+├── client
+│   └── ayon_<addon_name>
```
iLLiCiTiT, line 5:
```suggestion
│   └── <client addon name>  # Example: ayon_core
```

ayon_server/addons/rezbuild.py
```diff
@@ -0,0 +1,112 @@
...
+└── server
+
+Other folders might or might not be present, but they are essentially ignored.
+
+Will zip up the `client/ayon_<addon_name>` and place it into `private/client.zip`
+and copy the contents of `server` into the root of the package.
+
+<package root>
+├── addon
+│   ├── frontend
```
iLLiCiTiT, line 16:
```suggestion
├── frontend
├── server
```

ayon_server/addons/rezbuild.py
```diff
@@ -0,0 +1,112 @@
...
+│   ├── __init__.py
+│   ├── private
+│   │   ├── client.zip
+│   │   └── pyproject.toml
+│   ├── settings
+│   └── version.py
+├── package.py
+└── build.rtx
```
iLLiCiTiT, line 24:
```suggestion
├── frontend
│   ├── dist
├── server
│   ├── __init__.py
│   └── settings
├── private
│   ├── client.zip
│   └── pyproject.toml
├── package.py
└── build.rtx
```

ayon_server/addons/rezbuild.py
```diff
@@ -0,0 +1,112 @@
...
+
+This expects a certain folder structure:
+├── client
+│   └── ayon_<addon_name>
+├── package.py
+└── server
+
+Other folders might or might not be present, but they are essentially ignored.
+
+Will zip up the `client/ayon_<addon_name>` and place it into `private/client.zip`
```
iLLiCiTiT, line 11:
```suggestion
Will zip up the `client/<client module name>` and place it into `private/client.zip`
```

ayon_server/addons/rezbuild.py
```diff
@@ -0,0 +1,112 @@
...
+            )
+
+
+def install_addon():
+    server_dir = SRC_DIR / "server"
+
+    # Copy "server" to "addon"
+    if server_dir.is_dir():
+        print(f"Copying `server` directory contents to {DEST_DIR}")
+        _recursive_copy(server_dir, DEST_DIR)
```
iLLiCiTiT, line 60: Not what this does, but if there is `package.py` it does expect to have server code in `server` subfolder.
```suggestion
        _recursive_copy(server_dir, DEST_DIR / "server")
```

PR #49, opened 2023-09-18
emits event on project creation

tweak-wtf:


martastain (approved):
Thanks for the patch. I am not 100% convinced the event should be dispatched from here - in other cases, we emit them from endpoints, to have all the data available (namely user name, sender), but this does the trick.  

On the other hand - addons may want to use create_project_from_anatomy directly a custom API endpoints and in that case it would be more convenient to have it here.

PR #56, opened 2023-10-26
Prefix folder path with a slash

martastain:
API (both rest and graphql) now returns folder paths prefixed with a slash (as in unix absolute paths). 

It is a backward compatible change as everywhere API ACCEPTS a folder path (graphql filters, access control), it allows using both prefixed and unprefixed versions.


ayon_server/entities/folder.py
```diff
@@ -227,7 +234,7 @@ async def commit(self, transaction=None) -> None:
         cache: dict[str, dict[str, Any]] = {}
 
         async for row in Postgres.iterate(query, transaction=transaction):
-            parent_path = "/".join(row["path"].split("/")[:-1])
+            parent_path = "/".join(row["path"].strip("/").split("/")[:-1])
```
Minkiu, line 237: Maybe we could do the strip above and use the new var later on too:
```py
row_path = row["path"].strip("/")
parent_path = "/".join(row_path.split("/")[:-1]
...
cache[row_path] = attr
```

Minkiu:
I know this is beyond the scope of this PR, but wondered if it might be worth using `pathlib.Path` for the `path` (not in the db, but when handling `row['path']`?

We could then do stuff like:
```py
from pathlib import Path
...
# row["path"] = "/assets/characters/xenomorph"
row_path = Path(row["path"])
row_path.parent # PosixPath('/assets/characters')
row_path.parent.relative_to("/") # PosixPath('assets/characters')
row_path.relative_to("/") # PosixPath('assets/characters/xenomorph')
row_path.as_posix()  # '/assets/characters/xenomorph'
```

mkolar:
I believe this has now bee rigorously tested and can be merged now.

mkolar (approved):

PR #68, opened 2023-11-13
More options for settings enum resolvers

martastain:
Enum resolvers now can accept source addon itself and use its own settings

```python
async def recursive_enum_resolver(
    addon: "BaseServerAddon",
    settings_variant: str = "production",
    project_name: str | None = None,
) -> list[str]:
    if addon is None:
        return []

    if project_name:
        settings = await addon.get_project_settings(project_name=project_name, variant=settings_variant)
    else:
        settings = await addon.get_studio_settings(variant=settings_variant)

    return settings.list_of_strings
```

Additionally resolvers accept `settings_variant` field (defaults to production, requires the latest develop frontend ) and built-in `folder_types_enum` and `task_types_enum` now return folder/task types from the primary anatomy preset, when invoked from studio settings.

### DIFF
--- a/ayon_server/settings/postprocess.py
+++ b/ayon_server/settings/postprocess.py
@@ -29,12 +29,8 @@ async def process_enum(
 ) -> tuple[list[str], dict[str, str]]:
     if context is None:
         context = {}
-    # else:
-    #     ctx_data = copy.deepcopy(context)
 
     resolver_args = inspect.getfullargspec(enum_resolver).args
-    # available_keys = list(ctx_data.keys())
-    print("Keys: ", resolver_args)
 
     ctx_data = {}
     for key in resolver_args:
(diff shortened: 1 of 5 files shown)


ayon_server/settings/postprocess.py
```diff
@@ -46,16 +46,17 @@ async def process_enum(
...
-            if type(item) is str:
-                enum_values.append(item)
-            elif type(item) is dict:
-                if "value" not in item or "label" not in item:
-                    logging.warning(f"Invalid enumerator item: {item}")
-                    continue
-                enum_values.append(item["value"])
-                enum_labels[item["value"]] = item["label"]
+    if not isinstance(enum, list):
+        return [], []
```
iLLiCiTiT, line 50:
```suggestion
        return [], {}
```

PR #82, opened 2024-01-10
Custom link names

martastain:
link table is changed to contain a custom `name` field.

The name is nullable and it may be used to use an integration-specific identifier. Links of the same type no longer need to be unique.

On the database level, the following has changed:

- Renamed `link_name`  to `link_type`
- Added `name`  column 
- Moving `author`  from `data` to its own column
- Remove `link_unique_IDX` 

API-wise, this PR should be 100% backward compatible, but `link` field in `[POST] /api/{project_name}/links` is now deprecated. Instead, please use `link_type` (for now, both may be used).

Graphql and rest both expose new optional `name` field.

### Testing

This PR is now in `ayon:experimental` image. In order to migrate existing projects to the new schema, the container must start (reloading only won't trigger the migration script).

If you restore an older project, restart the container afterward, so it is converted as well - projects with older schemas should work, only links will be broken.

link types weren't changed in any way.

### DIFF
--- a/api/links/links.py
+++ b/api/links/links.py
@@ -127,11 +127,14 @@ class CreateLinkRequestModel(OPModel):
 
     input: str = Field(..., description="The ID of the input entity.")
     output: str = Field(..., description="The ID of the output entity.")
-    link: str = Field(
-        ...,
-        description="The name of the link type to create.",
+    name: str | None = Field(None, description="The name of the link.")
+    link: str | None = Field(
+        None,
+        description="Link type to create. This is deprecated. Use linkType instead.",
         example="reference|folder|version",
     )
+    link_type: str | None = Field(None, description="Link type to create.")
+    data: dict[str, Any] = Field(default_factory=dict, description="Link data")
 
 
 @router.post("/projects/{project_name}/links")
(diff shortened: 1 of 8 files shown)


api/links/links.py
```diff
@@ -127,11 +127,14 @@ class CreateLinkRequestModel(OPModel):
...
-    link: str = Field(
-        ...,
-        description="The name of the link type to create.",
+    name: str | None = Field(None, description="The name of the link.")
+    link: str | None = Field(
+        None,
+        description="Link type to create. This is deprecated. Use linkType instead.",
         example="reference|folder|version",
     )
+    link_type: str | None = Field(None, description="Link type to create.")
```
Minkiu, line 136: MIght be worth having the `example` from `link` if they are the same.

In line 150 we'll fail if the `link_type` is not specified, so why allow `None` here?

iLLiCiTiT (approved):

ayon_server/graphql/resolvers/links.py
```diff
@@ -12,14 +12,20 @@ async def get_links(
...
+    first: int = 100,
+    after: str | None = None,
+    names: list[str] | None = None,
+    name_ex: str | None = None,
 ) -> LinksConnection:
     project_name = root.project_name
 
     edges: list[LinkEdge] = []
 
+    print("names", names)
```
Minkiu, line 25: remember to remove these

PR #87, opened 2024-01-24
Force folder delete flag

martastain:
Allow deleting folders with already published products/versions/representation either using `?force=true` flag in `[DELETE] /api/projects/{project_name}/folders/{folder_id}` or by setting `force=True` in operation data, eg:

```json
{
  "operations": [
    {
      "id": "123456789132456789123456789",
      "type": "delete",
      "entityType": "folder",
      "entityId": "132456aaa13246a464a897a2",
      "force": true
    }
  ]
}
```


iLLiCiTiT:
I think that the idea is that force should also delete all children folders, which is not handled with this PR. Or that will be handled in  frontend?

martastain:
If a folder has subfolders, they are deleted recursively by default. That is handled by `parent_id UUID REFERENCES folders(id) ON DELETE CASCADE` in the folders table. The same works for tasks and workfiles, but there's a strict constraint that disallows deleting folders (even nested ones) that have a product published. When a product is deleted, it in cascade deletes versions and representations, so the only constrained point is folder-product, which is by design.

iLLiCiTiT:
So, if I force delete `/Assets` which does not have anything published, it will also delete `/Assets/Bob` which has published stuff?

martastain:
> So, if I force delete `/Assets` which does not have anything published, it will also delete `/Assets/Bob` which has published stuff?

only if you set `force=true`. Otherwise, that constraint on `products.folder_id` will prevent it and the entire transaction rolls back.

martastain:
> So, if I force delete `/Assets` which does not have anything published, it will also delete `/Assets/Bob` which has published stuff?

errr.... no. only if you force delete `/assets/bob` directly. should it be possible to do it recursively? i think that would be quite dangerous.

iLLiCiTiT:
I think it should work that way in UI, does not have to work that way on backend.

mkolar:
Task linked: [OP-6932 Delete of parents should propagate to children](https://app.clickup.com/t/6658547/OP-6932)

mkolar:
@iLLiCiTiT is correct. This should allow wiping the whore hierarchy underneath

martastain:
force flag now causes recursive delete. Using it requires the user to be a manager

PR #89, opened 2024-02-05
Support for new addon structure

martastain:
The new addon structure makes server-side development easier as it is now possible to clone an addon repository directly to the addon storage, work on the code, and commit it back. Original `manifest.json` was replaced with `package.py` or `package.yaml` files (both variants are supported) in order to provide forward compatibility with the future Rez-based addons.

Example of an addon with the new structure is drafted at https://github.com/ynput/ayon-example-rez-addon 

package file is now the source of truth for the addon name and version, the rest of the metadata (title, frontend scopes...) are for now defined in the addon class as before. moving this to the package file is TBD.

package file also may also define required ayon version (`ayon_version = ">=1.03,<2.0.0"`) and this is respected in the marketplace.

if the addon is just cloned (not installed) and therefore `.git` folder exists in the addon directory, `+git` is appended to the version string, so it is possible to have a development version as well as a version installed from the resulting zip file on the same machine.


Minkiu (approved):
Haven't tested, but overall looks good.

PR #116, opened 2024-04-02
added __primary__ and __builtin__ anatomy presets

martastain:
Instead of the original `_` magic value, anatomy presets may be queried using `__primary__` and `__builtin__` 
original `_` is now alias to `__builtin__` to maintain backward compatibility.

Enum resolver for settings is available so it is possible to create fields such as:

```python
    anatomy_preset: str = Field(
        "__primary__",
        title="Anatomy preset",
        description="Anatomy preset to use",
        enum_resolver=anatomy_presets_enum,
    )
```

in that case, settings value will always point to the primary preset. 
If no primary preset is defined, it will point to the built-in.
User may then override the value with a an explicit preset name.

### DIFF
--- a/api/anatomy/__init__.py
+++ b/api/anatomy/__init__.py
@@ -143,6 +143,26 @@ async def set_primary_preset(preset_name: str, user: CurrentUser) -> EmptyRespon
     return EmptyResponse()
 
 
+@router.delete("/presets/{preset_name}/primary", status_code=204)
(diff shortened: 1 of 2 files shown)


jakubjezek001:
Does it mean that primary has to be set explicitly or it will be used as default behavior in case `ayon_api.create_project` is used with **preset_name** with **None**? Perhaps question rather for @iLLiCiTiT

jakubjezek001 (approved):

Issue #123, opened 2024-04-09
Requirements and compatibility definition for addon versions


martastain:
## Story

With the evergrowning list of Ayon addons and the fast pace of development, developers need a way
to declare which version of Ayon and other addons their addon is compatible with.

## Problem

Some addons may not work with certain versions of Ayon or other addons. In that case, the bundle
may not work as expected or may not work at all. This can lead to a bad user experience and a lot of
confusion.

Unfortunately REZ manifest don't provide a way to specify version constraints in the way that
we need, so we shoul aim to extend the manifest file to include this information.

## Proposal

Addon manifest (`package.py`) should contain the following information. 
Version compatibility is provided using "poetry" style version constraints (comma separated list of version ranges).

Checking will be performed using [semver.match](https://python-semver.readthedocs.io/en/2.9.0/api.html#semver.match)

Ayon already enforces semver versioning for addons and will use this information to check compatibility.

This won't break compatibility with existing addons, nor the REZ ecosystem.

### `ayon_server_version: str`

This key is already supported in the manifest file. It specifies the version of Ayon server that the addon is compatible with.

### `ayon_launcher_version: str`

This key specifies the version of the Ayon launcher that the addon is compatible with.

### `ayon_required_addons: dict[str, str]`

List of required addons and their version constraints. 
Bundle won't be created if any of the required addons are missing or incompatible.

### `ayon_compatible_addons: dict[str, str]`
List of compatible addons and their version constraints.
If an addon is present in the bundle, it should be compatible with the specified version range.

Bundle may be created even if some of the addons are not compatible, but the user will be warned.

## Example

```python

name = "example"
title = "Example addon"
version = "2.5.0"

plugin_for = ["ayon_server"]
build_command = ""

ayon_server_version = ">=1.0.8,<2.0.0"
ayon_launcher_version = ">1.5.0"

ayon_required_addons =  {
    "addon1": ">=0.1.0",
    "addon2": ">=0.1.0",
}
ayon_compatible_addons = {
    "addon3": ">=0.1.0",
    "addon4": ">=0.1.0",
}
```

martastain:
fair enough. we already use `ayon_version` key, but since it is not used that much, i think i could just make it an alias and for ayon_server_version, deprecate it and eventually start log warnings that it should be changed.

iLLiCiTiT:
> Bundle may be created even if some of the addons are not compatible, but the user will be warned.

[...]

PR #127, opened 2024-04-11
Addon compatibility check for bundles

martastain:
https://github.com/ynput/ayon-backend/issues/123

## Testing
- Create an addon with dependencies (as described in the linked issue)
- Try creating a bundle that does not meet the dependencies
- It should display an error

When the frontend counterpart of this PR is completed, the frontend will display errors and warnings inline.


martastain:
During bundle creation (when user edits the form), the frontend may call an endpoint with the same payload it would send for the actual create, but it would return sanity check result which may be displayed inline (and disable save button when checks fail).

image

martastain:
image

iLLiCiTiT:
> @iLLiCiTiT do you think create bundle endpoint should fail when there's an error or should i just disable the save option in the UI?

My personal opinion is that it should.

Does the information returned from server also report which addon caused the issue? E.g. `sitesync` is not installed, but who requested it?

martastain:
> Does the information returned from server also report what caused the issue? E.g. `sitesync` is not installed, but what requested it?

Yes. It is pretty granular. In this case (screenshot), sitesync is not requested by any of the addon, but it failed to start, hence this message - it references itself. (but it shouldn't be possible to select it at all). And every addon can return multiple issues.

martastain:
> > @iLLiCiTiT do you think create bundle endpoint should fail when there's an error or should i just disable the save option in the UI?
> 
> My personal opinion is that it should.

Sorry, but "it should" is quite ambiguous in this case :)

martastain:
I think that check could be performed by default and prevent creation if it fails, but we can have `force=true` argument to override if needed. That would be consistent with other use of `force` flag across the API

jakubjezek001:
I would like to test it, but the descriptions are missing some testing steps. Could we add those please, it will improve the review-ability?

martastain:
> I think that check could be performed by default and prevent creation if it fails, but we can have `force=true` argument to override if needed. That would be consistent with other use of `force` flag across the API

image

iLLiCiTiT:
> I think that check could be performed by default and prevent creation if it fails, but we can have force=true argument to override if needed. That would be consistent with other use of force flag across the API

I think we might need the force for dev bundles.

martastain:
> I think we might need the force for dev bundles.

Yep. That makes sense... I'll do that in https://github.com/ynput/ayon-frontend/pull/369

antirotor:
... and just to add to the bitching (but otherwise I  think the code looks solid) - it is missing PR in https://github.com/ynput/ayon-tests

iLLiCiTiT:
Works. We should add it to example addon and make nice frontend 🙂

iLLiCiTiT:
And maybe define example version ranges that are supported? E.g. `">=1.0.0,<1.1.0"` would work?

Innders:
```python
ayon_compatible_addons = {
    "tvpaint": "<=0.1.0",
}
```

```json
[
  {
    "severity": "warning",
    "addon": "webpublisher",
    "message": "tvpaint <=0.1.0 is required"
  }
]
```

Maybe the wording "required" should be slightly different here as this is a warning about compatibility not requirements.

Suggestion: `Webpublisher is only compatible with TVPaint <=0.1.0`

Even better would be human readable version `Webpublisher is only compatible with TVPaint version 0.1.0 or earlier.`

iLLiCiTiT:
> Maybe the wording "required" should be slightly different here as this is a warning about compatibility not requirements.

It is about requirement of a compatible version. It means "I don't require tvpaint to run, but if there is tvpaint I need version less than 0.1.0"

Innders:
> It is about requirement of a compatible version. It means "I don't require tvpaint to operate, but if there is tvpaint, I require version less or equal to 0.1.0"

But if it's a requirement then the severity=error.

Severity=warning should be like "It would be great if you had this version as we know it's definitely compatible but it's not required and you can create the bundle anyway but things might break btw".

Innders (changes requested):
Found a bug.

When webpublisher has a required addon the warning message will mention webpublisher not the required addon.

```python
# webpublisher 1.0.8-broken package.py
ayon_required_addons =  {
    "core": ">=0.1.0",
}
```

`[POST] api/bundles/check` request
```js
addons: {
...,
blender: "0.1.6",
core: null,
deadline: "0.1.9",
webpublisher: "1.0.8-broken"
...
}
```
Response

```json
{
    "success": false,
    "issues": [
        {
            "severity": "error",
            "addon": "webpublisher",
            "message": "webpublisher is required"
        },
    ]
}
```

Innders:
Found another bug.

```python
ayon_required_addons =  {
    "core": ">=0.1.0",
    "blender": ">=0.1.1",
    "fusion": None,
}
```

`"fusion": None` seems to break everything, the check will always return `success=true`.

Is it even possible to require to NOT have an addon. 

Maybe `cool-addon` would require that `uncool-addon` is not also installed.

Innders:
Is there some how we can include the required addon in the response?

```
    "issues": [
        {
            "severity": "error",
            "addon": "webpublisher",
            "message": "traypublisher 0.1.2 is required",
            "required": "traypublisher"
        },
    ]
    ```
    
    It would be handy to be able to click on the issue and it scroll you to the addon so you can change the version.

PR #131, opened 2024-04-12
Configurable event retention

martastain:
## Story

With the introduction of the activity feed, the event stream has primarily become a technical tool that no longer needs to be retained indefinitely. 

Since the activity feed collects data valuable for production, the event stream's value now lies in providing "undo" options and storing short-term technical data.

## Proposal

By using the AYON_EVENT_RETENTION_DAYS environment variable, the server can be configured to purge outdated events, thereby reducing database size and improving performance. 

By default, this feature will be turned off, but a reasonable default value could be approximately 90 days, which should suffice for most needs. Users can opt-in for automatic cleanup at any time.


jakubjezek001 (approved):
this looks good! can this be merged?

mkolar:
@martastain same question as Jakub.

PR #148, opened 2024-04-18
Filtering checklists in activity feed

martastain:
Adds an option to filter comments containing checklists.

That way, users can easily find comments that contain checklists, which are often used to track progress on tasks.

## Solution

- when a comment is created or updated, it is checked for the presence of a checklist
- if a checklist is found, boolean `hasChecklist` is stored to activity.data
- virtual `checklist` activityType was added to the activityTypes filter on activities resolver

When activityTypes filter contains `comment` it lists both normal comments and comments with checklists. 
When it contains `checklist` (and not comment) it lists only comments with checklists.

### DIFF
--- a/ayon_server/activities/create_activity.py
+++ b/ayon_server/activities/create_activity.py
@@ -6,7 +6,11 @@ from ayon_server.activities.models import (
     ActivityType,
 )
 from ayon_server.activities.references import get_references_from_entity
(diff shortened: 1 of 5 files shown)


Innders (approved):
Awesome! Tested on frontend and it all seems to be working 🥳

PR #150, opened 2024-04-18
Propagate publishes to activity feed

martastain:
Creating a new version now propagates to activity feed. activityData.context contains additional information to construct the message body.

image


martastain:
This is now availble in experimental, one version per task was published as POC in demo_Commercial

PR #174, opened 2024-05-09
Do not return empty site root overrides

martastain:
## PR Checklist

## Description of changes

When using `api/projects/{project_name}/roots` endpoint (GET/PUT), empty values (nulls or empty strings)
are no longer returned. When no override is present, exclude that root (or site) from the result.


kalisp:
`/roots` endpoint is returning empty dict for no overrides/removed overrides correctly.
image

But I would expect `siteRoots` to pick values from Anatomy (for particular site) and it doesn't:
image

kalisp (approved):
`siteRoots` returns expected values

PR #177, opened 2024-05-10
Add support for ayon_soft_required_addons

martastain:
## PR Checklist

## Description of changes

Addons may now provide `ayon_soft_required_addons` field, which specifies addons, which MUST be in the given version, if the addon is in a bundle. It is similar to `ayon_compatible_addons` with the difference this returns error instead of warning when the condition is not met.

### DIFF
--- a/api/bundles/check_bundle.py
+++ b/api/bundles/check_bundle.py
@@ -117,11 +117,11 @@ async def check_bundle(
 
         # Check for required addons.
         # If the required addon is not present, it's an error.
-        # If it is present, it must be compatible.
+        # If it is present, it must match soft_required.
+        # If it doesn't match compatibility, it's an warning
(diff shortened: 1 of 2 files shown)


api/bundles/check_bundle.py
```diff
@@ -158,6 +158,33 @@ async def check_bundle(
...
+            if b_version is None:
+                issues.append(
+                    BundleIssueModel(
+                        severity="error",
+                        addon=addon_name,
+                        message=f"{r_name} is required",
+                        required_addon=r_name,
+                    )
+                )
+            elif not is_compatible(b_version, r_version):
```
iLLiCiTiT, line 177:
```suggestion
            if not is_compatible(b_version, r_version):
```

iLLiCiTiT (approved):

PR #187, opened 2024-05-17
Activity feed: create user relations for version and folder entities

martastain:
## Description of changes

When an activity is created or updated, references to relevant users are created. This is useful to display relevant information to users within the project scope and later as the source of truth for inbox.

### Technical details

Relations are created for:

- folders: assignees on the task within the folder
- tasks: assignees
- versions: author, assignees on the linked task

More relations may be created in the future as this feature is extendable. Additionally, we should consider opting-out from creating the references as we figure out, which ones are really relevant to the users.

New concepts:

- `entityNames` filter in activities graphql resolver
- `role` field in activities.referenceData (can be `assignee` or `author`, but do not rely the value is present, since data may be from server version `<1.1.2` )

### Additional context

Example:

A comment was created by user `admin` on a folder. On that folder, there is a task assigned to `demouser76`.
Since a relation was created, in the following query we can see the comment as it may be relevant to the user:

GraphQL query:

```graphql
query UserReferences {
  project(name:"reftest"){
    activities(entityNames:["demouser76"], referenceTypes:["relation"], last:1){
      edges{
        node{
          activityType
          body
          referenceType
          referenceData
          activityData
        }
      }
    }
  }
}
```

Result:

```json
{
  "data": {
    "project": {
      "activities": {
        "edges": [
          {
            "node": {
              "activityType": "comment",
              "body": "Commenting on the folder",
              "referenceType": "relation",
              "referenceData": "{\"role\":\"assignee\"}",
              "activityData": "{\"files\":[],\"author\":\"admin\",\"origin\":{\"id\":\"1c0fceb0fccc11ee9e813be7463ccc84\",\"name\":\"folder\",\"type\":\"folder\",\"label\":\"folder\"}}"
            }
          }
        ]
      }
    }
  }
}
```

### DIFF
--- a/ayon_server/activities/create_activity.py
+++ b/ayon_server/activities/create_activity.py
@@ -77,17 +77,19 @@ async def create_activity(
 
     # Origin is always present. Activity is always created for a single entity.
 
-    references = [
+    references: set[ActivityReferenceModel] = set(extra_references or [])
+
+    references.add(
         ActivityReferenceModel(
             entity_id=entity_id,
             entity_type=entity_type,
             entity_name=None,
             reference_type="origin",
-        ),
-    ]
+        )
+    )
 
     if user_name:
-        references.append(
+        references.add(
             ActivityReferenceModel(
                 entity_type="user",
                 entity_name=user_name,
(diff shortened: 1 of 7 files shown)


Innders:
> Relations are created for:
> 
> * folders: assignees on the task within the folder
> * tasks: assignees
> * versions: author, assignees on the linked task

And this for all activity types "comment" "status.change" "assignee.add" "assignee.remove" "version.publish"?

Will you be able to limit it to just one activity type. 

For example:

Posting a new **_comment_** on a folder makes sense to notify all child task assignees but updating the **_status_** does not make sense.

Innders:
> * folders: assignees on the task within the folder

Will this play nicely when we have the addition of "reviewers" and "watches"?

Innders (approved):

martastain:
It is possible to filter out by activityType so we could collect the data and then just decide which to show.

martastain:
The logic tries to deduplicate references, so, in the case: "admin creates a comment on a version, which has author `demouser92`, which is also a single assignee on the linked task", the result is:

```
server-1    | <ActivityReference relation user demouser92>
server-1    | <ActivityReference origin version 406b929afcbc11ee9ea50242ac140004>
server-1    | <ActivityReference relation task 4064e576fcbc11ee9ea50242ac140004>
server-1    | <ActivityReference author user admin>
```

(relation to task is excluded, because `demouser92` is already referenced as the version author)

PR #199, opened 2024-05-21
REST API: Unify SiteID specifications 

martastain:
This PR unifies site_id dependency across rest endpoints. Two dependencies `SiteID` and `ClientSiteID` are used as specified in the linked issue.

### Discussion & future

Endpoints `/api/projects/{project_name}/roots/{root_name}` still use site_id provided in path. This is inconsistent, but probably not worthed breaking the backwards compatibility.

`/api/settings` still expect site id to be provided as a query argument. Should it also accept header option as it is called from the client application and site_id is by default sent with the request? or do we need granularity here and option to get just project settings? @iLLiCiTiT 

### Testing

- Save project site settings and ensure they are correctly stored when different sites are used
- Save different site overrides for project roots
- Ensure site overrides for settings and roots are correctly propagated to the client

### DIFF
--- a/api/addons/project_settings.py
+++ b/api/addons/project_settings.py
@@ -6,7 +6,7 @@ from nxtools import logging
 from pydantic.error_wrappers import ValidationError
 
 from ayon_server.addons import AddonLibrary
-from ayon_server.api.dependencies import CurrentUser, ProjectName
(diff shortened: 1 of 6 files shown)


antirotor (approved):
Thanks, it works for my use cases

PR #212, opened 2024-05-27
User inbox resolver and management

martastain:


Innders:
When calling `[POST] http://localhost:3000/api/inbox` does it update... `updatedAt`? It doesn't seem to.

PR #224, opened 2024-06-04
Activity Feed: comment auto suggestions

martastain:
Implements server-side resolving for entity autosuggestion in comments

### DIFF
--- a/ayon_server/suggestions/folder.py
+++ b/ayon_server/suggestions/folder.py
@@ -1,0 +1,110 @@
+from ayon_server.entities import FolderEntity
+from ayon_server.lib.postgres import Postgres
+
+from .models import (
+    FolderSuggestionItem,
+    TaskSuggestionItem,
+    UserSuggestionItem,
+)
+
+STYPE = list[UserSuggestionItem | TaskSuggestionItem]
+
+
+async def get_folder_suggestions(
+    user: str,
+    folder: FolderEntity,
+) -> dict[str, STYPE]:
+    """
+    Assignees: Every assignee in the project
+    Versions: Disabled - what versions would you want to see on a folder?
+    Tasks: Direct child tasks of the folder.
+    """
+
+    project_name = folder.project_name
+
+    result: dict[str, STYPE] = {"users": [], "tasks": []}
+
+    # get users:
+
+    query = f"""
+        WITH relevant_users AS (
+            SELECT unnest(t.assignees) as name
+            FROM project_{project_name}.tasks t
+        )
+
+        SELECT
+            u.name as name,
+            u.attrib->>'fullName' as label,
+            r.rel_count as has_task
+        FROM users u
+        LEFT JOIN LATERAL (
+            SELECT count(*) as rel_count
+            FROM project_{project_name}.tasks t
+            JOIN project_{project_name}.hierarchy h ON t.folder_id = h.id
+            WHERE
+                t.assignees @> ARRAY[u.name]
+                AND h.path LIKE '{folder.path.lstrip("/")}%'
+        ) r ON true
+        WHERE
+            u.name IN (SELECT name FROM relevant_users)
+        AND u.name != $1
+        ORDER BY
+            r.rel_count DESC,
+            u.name;
+    """
+
+    async for row in Postgres.iterate(query, user):
+        item = UserSuggestionItem(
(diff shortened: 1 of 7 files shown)


Innders:
Is it too crazy to accept `entityIds` (plural)? You can select more than one entity of the same type at one time.

martastain:
it is crazy... you are always commenting on one entity.

Innders:
> you are always commenting on one entity

Actually you can comment on multiple 😬 Okay let's leave it for now...

ayon_server/suggestions/models.py
```diff
@@ -0,0 +1,46 @@
...
+class ProductSuggestionItem(SuggestionItem):
+    id: str = Field(..., example="af3e4b3e-1b1b-4b3b-8b3b-3b3b3b3b3b3b")
+    name: str = Field(..., example="model_main")
+    product_type: str = Field(..., example="Model")
+    parent: FolderSuggestionItem | None = Field(None)
+
+
+class VersionSuggestionItem(SuggestionItem):
+    id: str = Field(..., example="af3e4b3e-1b1b-4b3b-8b3b-3b3b3b3b3b3b")
+    version: int = Field(..., example=1)
```
Innders, line 45: Could we also get `name` field here please

Innders (changes requested):

Innders (changes requested):
Users seems to be broken
https://github.com/ynput/ayon-backend/assets/49156310/50359cd1-77a6-4468-ae7f-175d889e65cd

ayon_server/suggestions/models.py
martastain: added

martastain:
> Users seems to be broken
> 
>  Screen.Recording.2024-06-04.at.16.34.30.mov

Are those users assigned to anything?

Innders:
> Are those users assigned to anything?

Looks like Demo User is the only user assigned to other tasks in the project. All the other users are not assigned to anything. 

If I add `hotel` user to any task on the project it will now appear in the search results.

I also think **me** should appear as well, I might want to tag myself.

Innders:
Status could influence relevance as well.

Status states ranked by most relevant:

1. In Progress
2. Blocked
3. Not started
4. Done

martastain:
> Status could influence relevance as well.
> 
> Status states ranked by most relevant:
> 
> 1. In Progress
> 2. Blocked
> 3. Not started
> 4. Done
> 

> I cannot use statuses in the query logic as they are configurable and may differ :(

Innders:
> > I cannot use statuses in the query logic as they are configurable and may differ :(

But status **states** do not change? 
image

martastain:
> > > I cannot use statuses in the query logic as they are configurable and may differ :(
> 
> But status **states** do not change? <img alt="image" width="1113" src="https://private-user-

That's true. Okay - it's doable, but I'm not convinced if it's worth another two subqueries. Let's keep that open as a "future extension" :)

ayon_server/suggestions/folder.py
```diff
@@ -0,0 +1,123 @@
...
+            u.attrib->>'fullName' as label,
+            r.rel_count as has_task
+        FROM users u
+        LEFT JOIN LATERAL (
+            SELECT count(*) as rel_count
+            FROM project_{project_name}.tasks t
+            JOIN project_{project_name}.hierarchy h ON t.folder_id = h.id
+            WHERE
+                t.assignees @> ARRAY[u.name]
+                AND h.path LIKE '{folder.path.lstrip("/")}%'
```
BigRoy, line 48: Does passing these strings in directly like this into the query mean we're now prone to SQL injection attacks?

martastain: project name is sanitized at the API level (in `ayon_server.api.dependencies`) and it won't allow any potentially malicious characters.

folder path is auto-generated as a materialized view from folder names, which also have strict rules on what characters could be used. Without handling queries as strings, I'd be f**** at many different places :)

Postgres does not allow passing table names as query arguments, so specifying schema as part of the query is necessary for all project-scoped FROM clauses.

Innders:
> That's true. Okay - it's doable, but I'm not convinced if it's worth another two subqueries. Let's keep that open as a "future extension" :)

Yeah cool, just throwing ideas around 👍

PR #226, opened 2024-06-06
Inbox: Websocket notification

martastain:
When a new activity is created, referenced users are notified using fire and forget websocket event with `inbox.message` topic.

The event uses the first line of the message body as its description `summary.isImportant` flag could be used to silence unimportant messages based on the user preferences.


Innders (changes requested):
It seems that a user will receive the message even if the new inbox item is not relevant them and wouldn't end up in their inbox. We are receiving all messages for everyone.

Innders (approved):

Issue #236, opened 2024-06-12
GraphQL: Normal users can list users from projects they are not assigned to


martastain:
### To reproduce

- Be a normal user
- Know a project name but don't have access to it
- Go to GraphQL explorer
- List users on another project

![image](https://github.com/ynput/ayon-backend/assets/5007200/9c689cb1-0796-4a2e-b25c-18685a408742)

### Expected behavior

User shouldn't be permitted to see users on other projects

PR #238, opened 2024-06-12
Disallow listing users from unassigned projects

martastain:
Fixes: https://github.com/ynput/ayon-backend/issues/236

It is no longer possible to list users from project that you are not assigned to.
Additionally, a status code from AyonException is passed to graphql error extensions.

### DIFF
--- a/ayon_server/graphql/resolvers/users.py
+++ b/ayon_server/graphql/resolvers/users.py
@@ -75,6 +75,9 @@ async def get_users(
         cnd = f"({cnd1} OR {cnd2} OR {cnd3})"
         sql_conditions.append(cnd)
 
+    # TODO: allow listing users from all project for normal users, but only if they are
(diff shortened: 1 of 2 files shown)


ayon_server/graphql/resolvers/users.py
```diff
@@ -64,13 +64,20 @@ async def get_users(
         sql_conditions.append(f"users.name IN {SQLTool.array(names)}")
 
     if project_name is not None:
+        if not user.is_manager:
```
iLLiCiTiT, line 67: Is `is_admin` the same as `is_manager`?

martastain: is_admin implies is_manager, so if the user is admin or a service, it evaluates as true as well

iLLiCiTiT (approved):

PR #243, opened 2024-06-17
Exclude addon endpoints and auth headers from openapi

martastain:
This PR introduces changes to the way we use OpenAPI schema for generating TypeScript definitions.

### Addon Endpoints Filtering

Addon endpoints, which are primarily relevant for addon developers, will now be excluded from the schema by default.
These endpoints can be included by setting the `ayon_openapi_include_addon_endpoints` environment variable.
This change reduces unnecessary noise in the schema and improves clarity for the majority of users.

### Authorization Options Streamlining

Shared authorization options, which are repeated across all endpoints, are now excluded from the schema.
Instead, authorization methods are comprehensively described in the Authorization section of the REST documentation. This avoids redundancy and centralizes authorization information for easier maintenance and reference.

### Security Considerations

Excluding addon endpoints by default enhances operational security as these endpoints could potentially expose sensitive information or vulnerabilities. Opting in to include these endpoints ensures they are only exposed when necessary.

### DIFF
--- a/ayon_server/api/dependencies.py
+++ b/ayon_server/api/dependencies.py
@@ -33,9 +33,11 @@ from ayon_server.utils import (
 
 
 async def dep_access_token(
-    authorization: Annotated[str | None, Header()] = None,
-    token: Annotated[str | None, Query()] = None,
-    access_token: Annotated[str | None, Cookie(alias="accessToken")] = None,
+    authorization: Annotated[str | None, Header(include_in_schema=False)] = None,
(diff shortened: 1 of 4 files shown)


Innders (approved):
I can confirm that it has removed all the addon endpoints as my codegen has removed all the addon endpoints and does not error when trying to generate the sitesync addon endpoint (because it's not there).

image

I also noticed that the headers and cookies and token param has been remove 👍 

image

PR #249, opened 2024-06-19
Server-side kanban resolver

martastain:
Provides a graphql resolver for Kanban data using an optimized query.

image

Most of the fields should be straightforward, `lastVersionWithThumbnailId` is used when the task does not have its own thumbnailId - in that case, id of the last (or hero) version that has ID is returned in this field to be used instead.

### projects : list[str], optional
List of project IDs to filter tasks. If not specified, tasks from all projects are listed. For non-managers, the result is limited to projects the user has access to. Inactive projects are never included.

### assignees : list[str], optional
List of user names to filter tasks. If the invoking user is a manager, tasks assigned to the specified users are listed. If not provided, all tasks are listed regardless of assignees. For non-managers, this is always set to [user.name].


Innders (changes requested):
image

`
2024-06-19 11:45:01 ERROR      server          GraphQL: /usr/local/lib/python3.11/site-packages/graphql/execution/execute.py:774 (kanban/edges/0/node/dueDate) 'str' object has no attribute 'isoformat'
`

martastain:
> `2024-06-19 11:45:01 ERROR server GraphQL: /usr/local/lib/python3.11/site-packages/graphql/execution/execute.py:774 (kanban/edges/0/node/dueDate) 'str' object has no attribute 'isoformat'`

Thanks! fixed.

PR #253, opened 2024-06-20
Fix: saving deeply nested settings cancels-out the original values

martastain:


ayon_server/settings/overrides.py
```diff
@@ -169,42 +169,66 @@ def extract_overrides(
     explicit_pins: list[list[str]] | None = None,
     explicit_unpins: list[list[str]] | None = None,
 ) -> dict[str, Any]:
+    existing_overrides = {**existing} if existing else {}
```
BigRoy, line 172: Is this just trying to make a copy?
```suggestion
    existing_overrides = existing.copy() if existing else {}
```

Also be aware that this will not be a `deepcopy` as such the input may still get affected.
So maybe this is safer:
```suggestion
    existing_overrides = copy.deepcopy(existing) if existing else {}
```

ayon_server/settings/overrides.py
```diff
@@ -169,42 +169,66 @@ def extract_overrides(
...
+        original_object: BaseSettingsModel,
+        new_object: BaseSettingsModel,
+        existing_overrides: dict[str, Any],
+        target: dict[str, Any],
+        path: list[str],
+    ):
+        for field_name in original_object.__fields__.keys():
+            old_child = getattr(original_object, field_name)
+            new_child = getattr(new_object, field_name)
+            rpath = [*path, field_name]
```
BigRoy, line 186: What's `rpath` variable supposed to mean? `right_path`, `real_path`? `relative_path`? `recursive_path`? Can we get a better name for it so the code is easier to grasp?

ayon_server/settings/overrides.py
```diff
@@ -169,42 +169,66 @@ def extract_overrides(
     explicit_pins: list[list[str]] | None = None,
     explicit_unpins: list[list[str]] | None = None,
 ) -> dict[str, Any]:
+    existing_overrides = {**existing} if existing else {}
+
     result: dict[str, Any] = {}
```
BigRoy, line 175: Looking at the logic and usage below it may be good to already here cache pins and unpins into a `set` for faster lookups.

```suggestion
    explicit_pins = set(explicit_pins) if explicit_pins else set()
    explicit_unpins = set(explicit_unpins) if explicit_unpins else set()
```

It would also make the code easier to read below because you don't need the `(explicit_pins or [])` logic there.

ayon_server/settings/overrides.py
BigRoy: Nevermind - this seems redundant copy because we're not editing the `existing_overrides` down below - we're solely using it as lookup. Which makes me wonder why we need the `{**existing}`?

ayon_server/settings/overrides.py
```diff
@@ -169,42 +169,61 @@ def extract_overrides(
...
-    crawl(default, overriden, existing_overrides, result)
+    crawl(default, overriden, existing_overrides, result, [])
+
+    # remove paths that are explicitly unpinned
+
+    for path in explicit_unpins or []:
+        current = result
+        for key in path[:-1]:
+            current = current[key]
+        current.pop(path[-1], None)
```
BigRoy: This may be easier to read with a dedicated function:
```python
def pop_path(data: dict, path: List[str]) -> Any:
    """Pop a child path nested deep in a dict (if it exists).

    If the key or any of its parents do not exist this will
    return None.

   Arguments:
       data (dict): The data to remove from.
       path (List[str]): The nested keys to pop.

    Returns:
        Any: The popped value, if any.
"""
    current = data
    parents, pop_key = keys.pop()
    for key in parents:
        current = data.get(key, {})
        if not current:
            return
        if not isinstance(current, dict):
            raise TypeError(f"Parent value for key {key} is not a dict: {current}")
    return current.pop(pop_key, None)
```

So that here the logic is just:
```python
for path in explicit_unpins:
    pop_path(result, path)
```

martastain: i realized, there already is a function to do this in ayon_server.utils :D

ayon_server/settings/overrides.py
martastain: field_path it is

ayon_server/settings/overrides.py
BigRoy: Even better!

ayon_server/settings/overrides.py
martastain: explicit pins and unpins need to be list of lists as path is a list, which is unhashable. each path could be converted to set as well, but there are just too many places where path is list[str]....

ayon_server/settings/overrides.py
```diff
@@ -169,42 +169,59 @@ def extract_overrides(
     explicit_pins: list[list[str]] | None = None,
     explicit_unpins: list[list[str]] | None = None,
 ) -> dict[str, Any]:
-    result: dict[str, Any] = {}
+    existing_overrides = existing or {}
+    explicit_pins = explicit_pins or []
+    explicit_unpins = explicit_unpins or []
```
BigRoy, line 174: I noticed you kept these `lists` instead of `set` likely because the paths you're putting in are `list` and therefore not hashable. However, if we'd make those `tuple` they are hashable and we can get `O1` performance on the lookup?

ayon_server/settings/overrides.py
martastain: yep. it was originally modified, but it is no longer a case. fixed

ayon_server/settings/overrides.py
BigRoy: Ah you seem to have commented on this [here](https://github.com/ynput/ayon-backend/pull/253#discussion_r1647372643) already. Ignore me if you don't care. I'm not sure how large these lookup lists tends to be in production use cases so it's hard to say whether it's worth the performance diffs.

martastain: Usually.... zero. lower 10s tops during settings copy :)

PR #257, opened 2024-06-28
Activity feed: notify clients when an activity is created, updated or deleted

martastain:
When an activity is created RT `activity.created` event is dispatched with a list of all affected entities. that should trigger cache invalidation of the activity feed.

image

Keep in mind, that sender matching will prevent originating browser tab receiving the message, so whoever creates the activity is responsible to invalidate their own cache. 

Don't forget subscribing to `activity.created` topic.

(the same structure is used for `activity.updated` and `activity.deleted` topics

### DIFF
--- a/api/activities/activity.py
+++ b/api/activities/activity.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 
-from fastapi import BackgroundTasks
+from fastapi import BackgroundTasks, Header
 
 from ayon_server.activities import (
     ActivityType,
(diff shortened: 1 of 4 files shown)


Innders:
A user can receive a message from a project they do not have access to.

image

martastain:
Hmmm. True, but there's not much they can do with that. As the message does not contain any data, just a bunch of IDs. 

It would be nice to filter out messages by access rights tho. I could do that in the websocket module as i hold metadata of connected users. But i think that's out of the scope of this PR. creating a second issue for it.

Innders:
> As the message does not contain any data

projectName is sensitive

martastain:
Oh. Yes. I'll start working on it.

PR #271, opened 2024-07-16
Use strict models for marketplace endpoints

martastain:
image


Innders (changes requested):
missing `marketAddonList` models :)

martastain:
Sorry. added :)

PR #273, opened 2024-07-16
Use CTE for enroll event resolver

martastain:
Judging by preliminary tests, this should speed-up enroll endpoint at least 40x with large events tables.

### DIFF
--- a/ayon_server/events/enroll.py
+++ b/ayon_server/events/enroll.py
@@ -153,14 +153,13 @@ async def enroll_job(
             # continue to the next event
             if sequential:
                 return None
-            continue  # implicit. for readability
+            continue
 
-        else:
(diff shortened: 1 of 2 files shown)


ayon_server/events/enroll.py
```diff
@@ -133,24 +134,33 @@ async def enroll_job(
...
+                depends_on=row["source_id"],
+                user=user_name,
+                description=description,
+                finished=False,
+            )
+
+        except ConstraintViolationException:
+            # for some reason, the event already exists
+            # most likely because another worker took it
+
```
iLLiCiTiT, line 151: You could just set `new_id = None` under exception handling, and keep resto of the code as was.

PR #287, opened 2024-07-19
Anatomy: Entity type scoped statuses

martastain:
Exposes "scope" field in project anatomy (list of strings) that allows limiting statuses to certain entity types.

If the field is empty, status will be available for all entities. Otherwise only on specified types (folder / task / product / version / representation / workfile ).

image

This need a counterpart implementation in the frontend (and the launcher)


jakubjezek001 (approved):
this is looking good! this will help allign statuses with Flow (SG)

iLLiCiTiT (approved):
It is ok. This might require minor version bump.

mkolar:
Is anything preventing this from being merged?

martastain:
> Is anything preventing this from being merged?

It is not implemented in the frontend, so changes here won't affect anything yet. But i guess we can merge it and implement the frontend part later.

PR #289, opened 2024-07-22
Always return dict for entity.data

martastain:
Return `"{}"` instead of null for empty entity.data fields in graphql.

image

In REST, it seems it already wored fine:

image

Please consider if this is a desired behavior - since in graphql it is necessary to json.loads the data field, using "null" would allow skip this and gain a little performance (it is much easier to validate `is None` than parsing the json string)


martastain:
Closing - handled in ayon-python-api: https://github.com/ynput/ayon-python-api/pull/166

PR #291, opened 2024-07-23
Reviewables sorting and patching

martastain:
Adds patch endpoints for sorting and relabeling reviewables

image

image


Innders (approved):
Works great!!!!

martastain:
Don't mind the long label (even tho `text-overflow: ellipsis` would be nice :) ), but i'd prefill the InputText with the original name and pre-select it - that way, you would be able to start typing to overwrite it completely ,or just press an arrow key to unselect the text and edit the existing label 

image

PR #304, opened 2024-08-01
Resolve: fix version argument parsing

martastain:


api/resolve/__init__.py
```diff
@@ -229,11 +135,18 @@ def get_product_conditions(product_name: str | None) -> list[str]:
...
+
     if version_name == "*":
         return []
+
     if version_name.startswith("v"):
         version_name = version_name[1:]
+
+    if version_name.isdigit():
+        # strip leading zeros
+        version_name = version_name.lstrip("0")
```
BigRoy, line 147: This stripping would technically be redundant:
```python
>>> int("00028")
28
```
(Py 3.9.13)

```suggestion
```

api/resolve/__init__.py
```diff
@@ -243,9 +156,11 @@ def get_version_conditions(version_name: str | None) -> list[str]:
             )
         """
         ]
+
     if version_name == "hero":
         return ["v.version < 0"]
 
+    logging.debug(f"Invalid version name: {version_name}")
```
BigRoy, line 163: Due to logic above - if the input value would've been `vv001` it would strip the first v, making the reported invalid version name `v001`  even though the input value was `vv001`. We should maybe make sure we keep a copy of the input value and make sure to log the source input value.

BigRoy (commented):
Yes, nice - some small comments! 

I've also changed your PR description to include some information.

api/resolve/__init__.py
```diff
@@ -229,11 +135,18 @@ def get_product_conditions(product_name: str | None) -> list[str]:
...
+
     if version_name == "*":
         return []
+
+    original_version_name = version_name
     if version_name.startswith("v"):
         version_name = version_name[1:]
+
+    if version_name.isdigit():
+        # strip leading zeros
```
BigRoy, line 147: Comment is not relevant anymore
```suggestion
```

BigRoy (approved):
Don't know how to test this PR - but the code changes make sense.

BigRoy:
Tagging @mkolar @dee-ynput for visibility to double check whether `v=1` should be valid inside a ayon entity URI as opposed to requiring the version name to be prefixed with `v`.

With this PR `version=1` and `version=v001` (and `version=001`) works the same way. Previously only `v001` would work.

PR #305, opened 2024-08-02
Settings: Improved SettingsField

martastain:
`SettingsField` is no longer just an alias to `pydantic.Field` but reimplements it and provides type checking for all additional types used in settings. **kwargs is now used just for backward compatibility (preventing raising errors, instead it logs a message about unsupported argument passed)

image

### DIFF
--- a/ayon_server/settings/settings_field.py
+++ b/ayon_server/settings/settings_field.py
@@ -36,7 +36,7 @@ def SettingsField(
     regex: str | None = None,
     discriminator: str | None = None,
     repr: bool = True,
-    # Ayon settings specifics
+    # AYON settings specifics
     example: Any = None,
     enum_resolver: AnyCallable | None = None,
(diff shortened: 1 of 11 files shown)


ayon_server/settings/settings_field.py
```diff
@@ -0,0 +1,125 @@
...
+    min_items: int | None = None,
+    max_items: int | None = None,
+    unique_items: bool | None = None,
+    min_length: int | None = None,
+    max_length: int | None = None,
+    allow_mutation: bool = True,
+    regex: str | None = None,
+    discriminator: str | None = None,
+    repr: bool = True,
+    # Ayon settings specifics
```
iLLiCiTiT, line 39:
```suggestion
    # AYON settings specifics
```

iLLiCiTiT (approved):

PR #307, opened 2024-08-02
Resolver: Handle malformed URIs and provide error messages

martastain:
When a provided URI is malformed or does not make sense, return "error" field with the problem description instead of empty entity list.

When a URI issue is recoverable (such as "cAsEiNsEnSiTiVe" version name), attempt to fix the problem

```json
[
  {
    "uri": "ayon+entity://ayontest/asset/char_hero?product=usdAsset&version=v00000021&representation=usd",
    "entities": [
      {
        "projectName": "ayontest",
        "folderId": "ac991b60b4cb11eeb7fc7730318c5a70",
        "productId": "506507b5ec8b11ee86565feb1905149c",
        "versionId": "d21458da394211ef8d7e0b740ae1b9fa",
        "representationId": "d21ee02e394211ef8c180b740ae1b9fa",
        "filePath": "{root[work]}/ayontest/asset/char_hero/publish/usd/usdAsset/v021/ynts_char_hero_usdAsset_v021.usd",
        "target": "representation"
      }
    ]
  },
  {
    "uri": "ayon+entity://ayontest/asset/char_hero?product=usdAsset&version=lAtEsT&representation=usd",
    "entities": [
      {
        "projectName": "ayontest",
        "folderId": "ac991b60b4cb11eeb7fc7730318c5a70",
        "productId": "506507b5ec8b11ee86565feb1905149c",
        "versionId": "1aa7484c4b5711efab716f26fd95f971",
        "representationId": "1ab343554b5711efb2fc6f26fd95f971",
        "filePath": "{root[work]}/ayontest/asset/char_hero/publish/usd/usdAsset/v025/ynts_char_hero_usdAsset_v025.usd",
        "target": "representation"
      }
    ]
  },
  {
    "uri": "ayon+entity://ayontest/asset/char_hero?product=usdAsset&version=vr0025&representation=usd",
    "error": "Invalid version name: vr0025"
  },
  {
    "uri": "ayon+entity://ayontest/asset/char_hero?product=usdAsset&version=v0025&task=modeling",
    "error": "Tasks and workfiles cannot be queried with products"
  },
  {
    "uri": "ayon+entity://ayontest/asset/char_hero?product=bambul\u00e1k",
    "error": "Invalid name: bambul\u00e1k"
  },
  {
    "uri": "ayon+entity://k\u0148our/flabadob/blib/blub/blip",
    "error": "Invalid project name: k\u0148our"
  },
  {
    "uri": "ayon+entity://ayontest/chrrr!",
    "error": "Invalid path element: chrrr!"
  }
]
```

### DIFF
--- a/api/resolve/__init__.py
+++ b/api/resolve/__init__.py
@@ -4,7 +4,6 @@ from typing import Any
 from urllib.parse import parse_qs, urlparse
 
 from fastapi import APIRouter, Query
-from nxtools import logging
 
 from ayon_server.api.dependencies import ClientSiteID, CurrentUser
 from ayon_server.exceptions import BadRequestException, ServiceUnavailableException
@@ -24,6 +23,7 @@ router = APIRouter(tags=["URIs"])
 
 
 SDF_REGEX = re.compile(r":SDF_FORMAT_ARGS.*$")
+NAME_VALIDATOR = re.compile(NAME_REGEX)
 
 
 def sanitize_uri(uri: str) -> str:
@@ -37,8 +37,8 @@ def validate_name(name: str | None) -> None:
         return
     if name == "*":
         return
-    name_validator = re.compile(NAME_REGEX)
-    assert name_validator.match(name), f"Invalid name: {name}"
+    if not NAME_VALIDATOR.match(name):
+        raise ValueError(f"Invalid name: {name}")
 
 
 def parse_uri(uri: str) -> ParsedURIModel:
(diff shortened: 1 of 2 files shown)


api/resolve/__init__.py
```diff
@@ -357,18 +362,28 @@ async def resolve_uris(
...
+                try:
+                    entities = await resolve_entities(
+                        conn,
+                        parsed_uri,
+                        roots.get(current_project, {}),
+                        site_id,
+                        path_only=path_only,
+                    )
+                except ValueError as e:
+                    result.append(ResolvedURIModel(uri=uri, error=str(e)))
```
BigRoy, line 386: For sake of consistency - should we still 'require' the `entities` field but force it empty? Or is that redundant data?

@Lypsolon might have opinions on that depending on how he uses the return value of `entities`?

BigRoy (approved):
Overall looks good to me - did have the one comment about the `entities` field now being optional and potentially being `None`. ⚠️

api/resolve/__init__.py
martastain: Yep. that would probably make sense from the compatibility point of view

BigRoy: Any reason to keep this PR lingering - or are you happy @Lypsolon ?

Lypsolon: well tomorrow is the release for ayon usd (the target)
and tbh i really dont want to just push code and hope. 

i believe the changes are good and i would like to spend time testing it with the cpp API and see what impact it has. 
but i did not yet have the time to test what real world impact it actually makes. 

this kind of testing against changes is hopefully done via CI after the next epic

martastain:
@Lypsolon could we merge this? It nows return an empty list instead of `null` when there's an error

Lypsolon:
> @Lypsolon could we merge this? It nows return an empty list instead of `null` when there's an error

i think we can merge it. 

i will say i believe recovering false URIs is not the best idea as it will further hurt the performance.

Lypsolon (approved):

PR #309, opened 2024-08-05
GraphQL: Added `hasReviewables` flag to folders, tasks and versions

martastain:


Innders (changes requested):
`hasReviewables` doesn't seem to work inside fragments. We have had this issue before with something else.

PR #310, opened 2024-08-05
Events: Reusable events

martastain:
This PR introduces an optional `reuse: bool = False` parameter to the `EventStream.dispatch` function. When set to True, if an event with the same hash already exists in the event stream, the existing event will be updated, and its ID will be returned instead of raising an exception.

This enhancement simplifies scenarios where previously it was necessary to first query the event by its hash and, only if it was not found, create a new one.

As a demonstration of this new functionality, the handling of the `addon.install` event has been refactored to leverage the reuse parameter.


ayon_server/events/eventstream.py
```diff
@@ -95,33 +99,71 @@ async def dispatch(
...
-                    "Event with same hash already exists",
-                ) from e
+                if reuse:
+                    raise ConstraintViolationException(
+                        "Unable to reuse the event. Another event depends on it",
+                    ) from e
+                else:
+                    raise ConstraintViolationException(
+                        "Event with same hash already exists",
+                    ) from e
```
iLLiCiTiT, line 166:
```suggestion
                raise ConstraintViolationException(
                    "Event with same hash already exists",
                ) from e
```

iLLiCiTiT (approved):
It works.

NOTE: I see advantage to know about all attempts to install addon? (Different issue and PR.)

PR #311, opened 2024-08-05
Settings: copy project settings when promoting a bundle

martastain:
When promoting a bundle from staging to production. All settings (site and project) should be copied during the process.

## Testing steps 

- Create a bundle and set it as staging
- Edit studio and project staging settings 
- Hit the rocket icon and confirm the dialog
- The bundle should be now be both in "production" and "staging" and the settings should be the same


m-u-r-p-h-y (changes requested):
I did the test, 

- studio overrides are transferred correctly
- project overrides are NOT transferred

we should also change the message to mention the project overrides 
image


also not sure if ending up with a bundle marked as Production and Staging at the same time is good. (but what else we should do . . .)
image

BigRoy:
> also not sure if ending up with a bundle marked as Production and Staging at the same time is good. (but what else we should do . . .)

This is actually totally fine - each would also still have their own settings so it's totally fine to share the bundle if you just want to test different settings in staging first.

BigRoy (approved):
So - this works perfectly on my end. I don't need to "hard refresh" either, it just works. (tested on windows + firefox).

There is an edge case visually where I do need to refresh and that is when I'm on an addon in studio settings currently that has staging overrides and I click the "promote bundle" button that sometimes the 'new' value doesn't display on the production bundle directly... then refreshing shows the right value. As if directly after refresh it's still getting the wrong value or whatever.

martastain:
Promote bundle needs to invalidate the frontend cache for the addon list as well... can do

PR #317, opened 2024-08-14
Removed exports from ayon_server.api to prevent circular imports

martastain:
Implicit exports from `ayon_server.api.__init__.py` are no longer supported. Only a few dependencies were exported using the old `dep_` style (without annotations), that are discouraged anyway.

It does not break any addons (nor the server itself), but reviewers please check whether there are no new errors in the server log during the server startup.

Bonus point: demogen should work now

Edit:

Waiting for:
 - https://github.com/ynput/ayon-jira/pull/6
 - https://github.com/ynput/ayon-sitesync/pull/34


MilaKudr (approved):
Screenshot 2024-08-15 130403

PR #319, opened 2024-08-15
Events: allow specifying source topics in list in enroll + batch event delete

martastain:
Enroll now accepts source topics both as strings with `*` wildcard or a list of strings. 

image

That maintains 100% backwards compatibility while allowing multiple topics selection without using expensive wildcards.

### Additional context

Additionally, this PR adds a new endpoint for batch updates of events. It was implemented in order to make the API testable, so it is just a minimal implementation only supporting event deletion, but it will be expanded later. 

image

### DIFF
--- a/api/events/enroll.py
+++ b/api/events/enroll.py
@@ -1,7 +1,13 @@
+import re
+
 from ayon_server.api.dependencies import CurrentUser
 from ayon_server.api.responses import EmptyResponse
 from ayon_server.events.enroll import EnrollResponseModel, enroll_job
-from ayon_server.exceptions import ForbiddenException, ServiceUnavailableException
+from ayon_server.exceptions import (
+    BadRequestException,
+    ForbiddenException,
+    ServiceUnavailableException,
+)
 from ayon_server.lib.postgres import Postgres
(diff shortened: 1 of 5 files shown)


api/events/enroll.py
```diff
@@ -62,21 +73,35 @@ async def enroll(
...
+    if not current_user.is_service:
+        raise ForbiddenException("Only services can enroll for jobs")
+
+    # source_topic
+    source_topic: str | list[str]
+
+    if isinstance(payload.source_topic, str):
+        source_topic = validate_source_topic(payload.source_topic)
+    else:
+        source_topic = [validate_source_topic(t) for t in payload.source_topic]
```
iLLiCiTiT, line 92: You could make your life easier with this change:
```suggestion
    if isinstance(payload.source_topic, str):
        source_topic = [validate_source_topic(payload.source_topic)]
    else:
        source_topic = [validate_source_topic(t) for t in payload.source_topic]
```

martastain: But I am saving several nanoseconds later when the topic is either evaluated as `topic = $1` in the case of string or `topic = ANY($1)` for lists in the SQL query :)

PR #321, opened 2024-08-19
Setup: Default access groups

martastain:
Provide a sane default access groups when the server is started for the first time and there are no explicit groups set in template.json

#### supervisor

Has unrestricted access to the project

#### artist 

Has unrestricted read/publish access to the project, but cannot change the project hierarchy

#### freelancer

Can read/publish the parts of the hierarchy where they have tasks assigned


m-u-r-p-h-y (approved):
default groups are created correctly

image

64qam (approved):

PR #323, opened 2024-08-20
GraphQL: Querying tasks in nested folders

martastain:
added `includeFolderChildren` to the tasks resolver, that expands the resultset with tasks under all children folders when `folderIds` list argument is used


Innders (approved):
Really cool and so fast!

PR #326, opened 2024-08-21
Settings: Migrate settings based on bundle

martastain:
## PR Checklist

- [x] migrate settings by bundle function
- [x] migrateSettingsByBundle endpoint
- [x] extend save bundle endpoint to allow settings migration in one step 
- [x] handle site settings properly
- [x] dispatch settings changed events
- [ ] use in-method caching of settings to reduce query amount
- [x] log everything

### Phase 1: Test migrateSettingsByBundle endpoint

- Turn off developer mode
- Create `OldBundle` bundle with an older version of an addon and change its studio and project settings
- Create `NewBundle` bundle with a newer version of the same addon and keep it untouched
- Set `OldBundle` to production and `NewBundle` to staging

Call (using postman, curl or httpie) a `[POST]` request with administrator or service privileges to `/api/migrateSettingsByBundle` with the following JSON payload

```json
{
  "sourceBundle": "OldBundle",
  "targetBundle": "NewBundle",
  "sourceVariant": "production",
  "targetVariant": "staging"
}
```

- Check the logs for errors. `Failed to validate XXXX with value XXX` warnings are expected if the target addon does not have migration method in place. that setting won't be migrated.
- Ensure the settings were correctly propagated in studio settings and project settings

- When https://github.com/ynput/ayon-backend/pull/311 is merged, use "promote bundle" function on the staging bundle and ensure site settings were also migrated


BigRoy:
> Check the logs for errors. Failed to validate XXXX with value XXX **warnings are expected if the target addon does not have migration method in place. that setting won't be migrated.**

Could you elaborate on this? Would it _never_ work if an addon doesn't specify a certain function? Or it would just be unable to resolve conflicts using use a migration method if version mismatches would have mismatching settings.

And when saying "mismatching settings". Is that, incompatible settings, like a type changing? Or if e.g. a setting was removed or added would it also fail?

martastain:
> Could you elaborate on this? Would it _never_ work if an addon doesn't specify a certain function? Or it would just be unable to resolve conflicts using use a migration method if version mismatches would have mismatching settings.
> 
> And when saying "mismatching settings". Is that, incompatible settings, like a type changing? Or if e.g. a setting was removed or added would it also fail?

If the model changes fields in an incompatible way such as `list[str]` to `str` and the newer addon does not implement  convert_settings_overrides method handling this change ( https://github.com/ynput/ayon-backend/blob/develop/ayon_server/addons/addon.py#L545 ) that field will be skipped during migration and the default will be used. 

It should however handle simpler changes such as `int` -> `str` and we could expand this implicit conversions based on experience. It uses the very same method that is used in the "copy settings dialog" in the frontend.

PR #328, opened 2024-08-29
Add unrealuasset as the publish template

moonyuet:
## PR Checklist

-   [ ] This comment contains a description of changes
-   [ ] Referenced issue is linked : https://github.com/ynput/ayon-unreal/issues/64 and  https://github.com/ynput/ayon-unreal/issues/66

## Description of changes
Added the published template setting for unreal uasset/umap publishing 
Related to this PR: https://github.com/ynput/ayon-core/pull/859

### Technical details
The current uasset/umap publisher is using default template name as the published path, but the default one would change the filename of the uasset/umap, which would corrupt the uasset/umap, and u sers cannot load the uasset/umap. There is some custom publish template needs to be added to make sure the uasset is published with original filename so that the file itself would not be corrupted.

### Additional context
The related ayon-core PR: https://github.com/ynput/ayon-core/pull/859
The related issues: https://github.com/ynput/ayon-unreal/issues/64 and  https://github.com/ynput/ayon-unreal/issues/66

### DIFF
--- a/ayon_server/settings/anatomy/templates.py
+++ b/ayon_server/settings/anatomy/templates.py
@@ -124,6 +124,11 @@ class Templates(BaseSettingsModel):
                 directory="{root[work]}/{project[name]}/{hierarchy}/{folder[name]}/publish/{product[type]}/{@version}",
                 file="{originalBasename}_{@version}.{ext}",
             ),
+            PublishTemplate(
+                name="unrealuasset",
+                directory="{root[work]}/{project[name]}/{hierarchy}/{folder[name]}/publish/{product[type]}/{product[name]}/{@version}",
+                file="{originalBasename}.{ext}",
+            ),
         ],
     )
 


LiborBatek:
@antirotor @iLLiCiTiT guys, who is the best for review of this one?

MustafaJafar:
> @antirotor @iLLiCiTiT guys, who is the best for review of this one?

Let me share my 2 cents.. 
Martin can push this to experimental ayond dev. and then guys who can test unreal PRs can use the experimental ayon and test https://github.com/ynput/ayon-core/pull/859 all together.

MustafaJafar:
I also wonder why anatomy settings live in ayon-backend repo.  
shouldn't it live in a designated addon e.g. ayon-core ?

dee-ynput:
That's actually a very legit question for which I would love to have the answer :)

martastain:
I don't know what do you mean. That core addon would somehow define the anatomy model? Or its part? Please elaborate how it would work.

Keep in mind that across the system we heavily depend on certain parts of the anatomy, such as statuses, folder/task types and roots.

MustafaJafar:
> Please elaborate how it would work.

tbh, It's hard for me to make a suggestion in this area.

But, I didn't expect that a PR to `ayon-backend`  is needed just to add some default settings.
Maybe it's a grey area.

Personally, I thought that `ayon-backend` includes the base classes definitions so that they are used somewhere else. e.g. `BaseServerAddon`.

martastain:
> But, I didn't expect that a PR to `ayon-backend` is needed just to add some default settings.
> Maybe it's a grey area.

This PR just adds a one item to defaults of the built-in anatomy preset. It might as well be an example. As soon a studio creates their own anatomy preset, they may delete this one. Maybe we could replace the default_factory method of publish templates with something else (pulling from an addon?). but is it worthed? the entire idea behind anatomy presets is that studio can create one or more custom preset they will use.

moonyuet:
> > But, I didn't expect that a PR to `ayon-backend` is needed just to add some default settings.
> > Maybe it's a grey area.
> 
> This PR just adds a one item to defaults of the built-in anatomy preset. It might as well be an example. As soon a studio creates their own anatomy preset, they may delete this one. Maybe we could replace the default_factory method of publish templates with something else (pulling from an addon?). but is it worthed? the entire idea behind anatomy presets is that studio can create one or more custom preset they will use.

We can mention in the documentation about this anatomy preset and how they can use it for the uasset publish. If the users get any question on the uasset publish, we can just guide them to the docs and ask them to add the anatomy preset and its relevant setting in the core addon.

MustafaJafar:
> We can mention in the documentation about this anatomy preset and how they can use it for the uasset publish. If the users get any question on the uasset publish, we can just guide them to the docs and ask them to add the anatomy preset and its relevant setting in the core addon.

So, instead of this PR, we will add something like the following in Docs:
1. For uasset publish, you'd need a custom anatomy template. 
2. Add the following template
image

I think this could work. But,  I'm not sure what is better product wise.

--- 

> Maybe we could replace the default_factory method of publish templates with something else (pulling from an addon?). but is it worthed?

maybe, I can't tell. 

--

tbh, I just didn't know that anatomy definitions are part of the backend itself. which sounds strange to me as I thought that is living in the core addon. which also not entirely correct idea because AYON server can have no addons and then we should have the anatomy definitions to be able to create any projects. 

anyways, I should leave it for some who is smarter and have wider overview than me.

dee-ynput:
The anatomy is a core concept of AYON, naming convention (template) is more or less a core concept, but DCC specific templates should definitely not be declared at server core.

That being said,:
- We have this kind of domain leakage in other places (application addon for example)
- We already broke the flow of domains with `maya2unreal`, `simpleUnrealTextureHero`, ...
- We have an Epic going on for the Settings UX which may help fix this soon.

So I would validate this PR and fix that concern (+more) later with a dedicated Epic 👍

martastain (approved):

PR #334, opened 2024-09-03
Resolve: Do not resolve all folders when uri contains just the project name

martastain:


Lypsolon (approved):
Basic Testing works.
changes also give me no reason why it should cause problems.

sooooo i Approve.

BigRoy:
Couldn't that resolve to the project entity instead? Or is that not a sensible entity to begin with?

martastain:
> Couldn't that resolve to the project entity instead? Or is that not a sensible entity to begin with?

It cannot. the result model is created for projectlevel entities (that have IDs) so it does not really makes sense. It would basically just for `ayon://myproject` return `entities:{"name": "myproject"}`

BigRoy (approved):
Fine by me. However, it always returning nothing kind of makes me feel we could just as well report an error - but hey, maybe empty responses is what everyone likes. 😎

PR #336, opened 2024-09-03
Unassign user from tasks upon deletion

martastain:
The linked issue has been opened for a while and i am not sure if there's a consensus whether this is a bug or feature. I've implemented it anyways, so maybe @mkolar or @dee-ynput could decide whether to merge it or not :)

### DIFF
--- a/ayon_server/entities/user.py
+++ b/ayon_server/entities/user.py
@@ -21,6 +21,7 @@ from ayon_server.exceptions import (
     NotFoundException,
 )
 from ayon_server.helpers.email import send_mail
+from ayon_server.helpers.project_list import get_project_list
 from ayon_server.lib.postgres import Connection, Postgres
 from ayon_server.lib.redis import Redis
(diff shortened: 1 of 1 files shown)


mkolar:
Yes I would unassign them upon deleting. But it does open another can of worms entirely....Should be allow deleting users at all? or rather shouldn't it be at least really hard? what happens with their comments, publisher, authored entities and so on

martastain:
I think we should have much bigger warning than this:

image

We already have this when user is renamed

image

Confirmation dialog should maybe contain a big big warning and something like "don't you rather want to disable the user instead" with a button to disable the user.

Innders:
Agreed.

1. The dialog should have a warning message like the rename users dialog. It could say:
```Deleting users can have unintended consequences. Would you rather deactivate them instead?```

2. [BONUS] The user must type the username into an input to confirm they are deleting the correct user.

image

PR #355, opened 2024-09-17
fix: column reference "id" is ambiguous

prOOrc:
If there is an ACL with the type - "Assigned", and there are tasks assigned to the user, there will be the "column reference id is ambiguous" exception when querying representations.


martastain (approved):
Thanks. Good catch!

PR #357, opened 2024-09-23
inherit setting field scope in list_overrides

martastain:


BigRoy (approved):
This seems to work.

**However!** it does seem like it's also 'copying' the "studio setting overrides" from another bundle as project setting overrides.

For example this:

image

Would be copied as:

image

So basically, if my studio settings do not match yet - those WOULD be copied over via project settings copy from another bundle.

Which I don't think I have faced before because I usually first transferred the "Studio Settings" and THEN the "Project Settings" and hence this may have already been the case. When doing a copy of the studio settings and then the project settings - this does remove the project overrides and leaves them 'inherited' again from the studio settings. So I guess it's fine.

Issue #380, opened 2024-10-10
Project files: Mark project directory as trashed upon project deletion


martastain:
We don't strictly want to delete all physical files from the storage when a project is deleted, but we need a simple way to list project directories that belong to deleted projects and be able to delete them either manually or automatically after a grace period

### Local filesystem

Upon project deletion, rename the project directory to `{project_name}.{timestamp}.trash` 

Since project names cannot contain `.` character, dot-delimination is non ambiguous, using a timestamp ( `int(time.time()`) allows multiple versions of the same project and restore the snapshot. 

Sufffixing the projectname with `.trash` makes it easy to filter & delete.

### S3

S3 does not support "directory" renaming so we need a different approach here:

- every project storage will use `{project_name}.{timestamp}` as the project root. where timestamp is the project created_at
- when project is deleted and recreated,  a new unique project root will be created
- clean-up process will delete project directories not attached to any project after a grace period.

PR #381, opened 2024-10-10
Project files: Trashing project files directories upon deletion

martastain:
https://github.com/ynput/ayon-backend/issues/380

### Local filesystem

Upon project deletion, the project directory is renamed to `{project_name}.{timestamp}.trash` 

Since project names cannot contain `.` character, dot-delimination is non ambiguous, using a timestamp ( `int(time.time()`) allows multiple versions of the same project and restore the snapshot. 

Sufffixing the projectname with `.trash` makes it easy to filter & delete.

> [!IMPORTANT] 
> This approach ensures 100% backward compatibility with existing projects

image

### S3

S3 does not support "directory" renaming so we need a different approach here:

- every project storage from now uses `{project_name}.{timestamp}` as the project root. where the timestamp is the project `created_at`, so it's deterministic, can be paired with the current version of the project and the list can be sorted by versions/periods
- when a project is deleted and recreated,  a new unique project root will be created
- The clean-up process will delete project directories not attached to any project after a grace period.

> [!WARNING]
> This is a breaking change that requires changing the project roots in S3 storages, 
> but as far as we know, S3 is not used anywhere in production yet.

image

### DIFF
--- a/api/projects/projects.py
+++ b/api/projects/projects.py
@@ -10,6 +10,7 @@ from ayon_server.exceptions import (
     ForbiddenException,
     NotFoundException,
 )
+from ayon_server.files import Storages
 from ayon_server.helpers.project_list import get_project_list
 from ayon_server.lib.postgres import Postgres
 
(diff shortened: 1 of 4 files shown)


BigRoy:
There is an S3 `mv` command (which apparently does `cp` and `rm` in the background according to comments [here](https://stackoverflow.com/a/26820022)). Are we avoiding the copy and remove just because that's very costly?

martastain:
Yep. it is costly and also time consuming - and it is not a good idea to do that as the background process:

- someone deletes a project
- moving is executed on the background
- during that period, a project with the same name is re-created (i did that yesterday several times)
- new project files are created in the same path and they are mixed with the old files

PR #396, opened 2024-10-18
Events: Optimize enrolls to avoid timeouts

martastain:
- [x] skip older events by default (but keep the option to include them)
- [ ] abort underlying SQL query in the case REST timeouts
- [ ] disallow executing the same request from the same sender multiple times at once
- [ ] pre-cache?


tanh609:
Thank you @martastain

PR #398, opened 2024-10-21
Watching and inbox rules improvements

martastain:
## Watching/unwatching a task

When watching a task:

- User gets added as a watcher to all versions of that task automatically.
- User gets added to any new versions of that tasks automatically.

When unwatching a task:

- User gets removed from watching all task versions.
- User does not get automatically added to new task versions.

### Note

Setting/unsetting a watcher should invalidate client cache as now the new state is only visible after page reload

## Status changes

Status changes are no loner considered important in the inbox, regardless user watching the entity or not

## get_inbox query

`get_user_inbox` query now takes in account whether the user has access to the project. If they're unassigned from the project, related items will disappear from the inbox (that also improves the performance).


Innders (approved):
Tested using the FE changes here: https://github.com/ynput/ayon-frontend/pull/893

PR #401, opened 2024-10-23
Fix attribute model for enums

martastain:
This pull request includes a small but important change to the `generate_model` method in the `ayon_server/entities/models/generator.py` file. The change ensures that if the `enum` field contains instances of `AttributeEnumItem`, their values are extracted correctly.


BigRoy (approved):
Makes sense. 👍 Didn't test. :)

PR #402, opened 2024-10-23
Handle empty response when returning cached 'nothing to do' from enroll

martastain:
This pull request includes a small but important change to the enroll endpoint. The change ensures that an `EmptyResponse` is returned if the cached result is not available.

### DIFF
--- a/api/events/enroll.py
+++ b/api/events/enroll.py
@@ -149,7 +149,7 @@ async def enroll(
 
         sloth("Request is already completed. Returning the cached result.")
         await Redis.delete("enroll", request_hash)
-        return cached_result.get("result")
+        return cached_result.get("result") or EmptyResponse()
 
     else:
         if cached_result:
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Seems fine to me.

Any reason you're not doing: `cached_result.get("result", EmptyResponse())` if for whatever the "result" is still empty in some other meaningful way?

Also, does `cached_result.get("result")` actually return a `EnrollResponseModel`? Because it seems to just be returning the JSON result from `Redis.get_json()` which looking at the code seems to `return json_loads` for example? However it could be that [this](https://github.com/ynput/ayon-backend/blob/develop/ayon_server/utils.py#L22) does indeed return that.

martastain:
> Any reason you're not doing: `cached_result.get("result", EmptyResponse())` if for whatever the "result" is still empty in some other meaningful way?

Yep, you're right. I'll clean it up

> Also, does `cached_result.get("result")` actually return a `EnrollResponseModel`? Because it seems to just be returning the JSON result from `Redis.get_json()` which looking at the code seems to `return json_loads` for example? 

Fastapi really doesn't care here. The result from orjson.loads is a plain dict, but it will work as long as it matches the response model.

PR #422, opened 2024-11-05
Fix: Saving staging directories to existing project anatomy

martastain:
- Fixes wrong key for staging directories when converting anatomy object to project data
- Adds extract_anatomy helper functions, that is needed from time to time in addons (namely Kitsu)


BigRoy (approved):
Works for me!

PR #432, opened 2024-11-12
Fix: Handle invalid thumbnails

martastain:


api/thumbnails/thumbnails.py
```diff
@@ -83,12 +84,17 @@ async def store_thumbnail(
     MAX_THUMBNAIL_WIDTH = 600
     MAX_THUMBNAIL_HEIGHT = 600
 
+    if guess_mime_type(payload) != mime:
+        raise BadRequestException("Mime type does not match the payload")
```
BigRoy, line 88: It could be nice - if this could report in the response what the mismatch was. Like what payload it received and what mime it expected?

PR #433, opened 2024-11-14
Warn when thumbnail mime mismatches instead of returning an error

martastain:
Publishing from the current version of maya uploads png thumbnail, but reports "image/jpeg" mime-type (or vice versa). That breaks publishing due to the strict check introduced in Ayon 1.5.5. Since we can determine the mimetype on the server from the payload headers, we just log the warning instead returning an error response.

### DIFF
--- a/api/thumbnails/thumbnails.py
+++ b/api/thumbnails/thumbnails.py
@@ -24,6 +24,7 @@ from ayon_server.exceptions import (
     BadRequestException,
     ForbiddenException,
     NotFoundException,
+    UnsupportedMediaException,
 )
 from ayon_server.files import Storages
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
This works.

During publishing nothing stands out nor is any warning logged (inside Maya).
However, the server logs show:
```
2024-11-14 13:50:29 2024-11-14 12:50:29 WARNING    server          Thumbnail mime type mismatch: Payload contains image/png but was requested to store image/jpeg
2024-11-14 13:50:29 2024-11-14 12:50:29 DEBUG      server          Resizing image from (1920, 1080) to (600, 336)
```

PR #435, opened 2024-11-14
Setup: Staged database migration

martastain:
This pull request changes the schema migration logic to avoid long transaction that could cause issues with larger databases.

* Old `schema.public.update.sql` was split to numbered files. That reduces transaction length.
* Introduced `db_migration` function to handle the database schema migration process, ensuring all migrations are applied in order.
* Modified the `main` function to use `db_migration` and save the current database version after applying migrations.

Since all migrations should be applied in order to handle old projects import, versioning will not be used just yet. But the DB will be prepared for it.

Since setup is crucial step, we now check whether it was executed properly and prevent the server from starting when an error occurs:

image

## Testing

Ensure setup finishes correctly

- on a fresh instance
- after container re-creation
- after running `make setup`

So far, it has been tested with existing local and new cloud instances:
image


setup/__main__.py
```diff
@@ -161,11 +154,14 @@ async def main(force: bool | None = None) -> None:
...
-    if bundle_data := DATA.get("initialBundle"):
-        if not isinstance(bundle_data, dict):
-            logging.warning("Invalid initial bundle data")
-        else:
-            await create_initial_bundle(bundle_data)
+        if bundle_data := template.get("initialBundle"):
+            if not isinstance(bundle_data, dict):
+                logging.warning("Invalid initial bundle data")
+            else:
+                await create_initial_bundle(bundle_data)
```
BigRoy, line 161: Is it correct that this changed in indentation?

martastain: yep. it should only happen during the initial setup. so when force_install is false, bundle data will be "None" anyways.

PR #437, opened 2024-11-15
Trigger event when entity parent changes

martastain:
Additional events are triggered when a top-level field of an entity is changed:

```python
{
    "parent_id": "parent_changed",  # for folders
    "folder_id": "folder_changed",  # for tasks and products
    "task_id": "task_changed",  # for workfiles and versions
    "product_id": "product_changed",  # for versions
    "version_id": "version_changed",  # for representations
}
```

image

Keep in mind that there are constraints in the API, that for example forbids changing hierarchy of folders (e.g. changing parent_id of a folder) when there's a published version in that folder.

### DIFF
--- a/ayon_server/events/patch.py
+++ b/ayon_server/events/patch.py
@@ -19,6 +19,11 @@ ADDITIONAL_COLUMNS = {
     "product_type": "product_type_changed",
     "author": "author_changed",
     "files": "files_changed",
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Missing `task_changed` for versions (they are not parents, but version have task relationship).

iLLiCiTiT (approved):
Works

PR #446, opened 2024-11-21
Links: Remove dead links upon entity deletion

martastain:
### Continuous clean-up

Upon project-level entity deletion, all its links should be removed. This is done by calling `ayon_server.helpers.entity_links.remove_entity_links` function (new)

### Deep clean-up

Clean-up procedure for remaining dead links could be done by calling `ayon_server.helpers.entity_links.remove_dead_links`.
This function is not yet exposed, but will become a part of a maintenance procedure.

Until then, it is possible to do that manually when needed:

1. Enter the python shell inside the server container: `docker compose exec server python`
2. Paste the following snippet and execute the script:

```python
import asyncio

from ayon_server.initialize import ayon_init
from ayon_server.helpers.entity_links import remove_dead_links

async def main():
    await ayon_init()
    await remove_dead_links("My_project")

asyncio.run(main())
```


jakubjezek001 (approved):
I have tested `remove_dead_links` function as described for container python script and it was working as expected. All my dead links are gone and GraphQL works now as expected. Thank you!

PR #487, opened 2025-01-17
GraphQL: return user levels as they are stored in the database

martastain:
Updated the `is_manager` assignment to depend solely on the `isManager` field from the data, rather than also considering the `isAdmin` and `isService` fields.

This makes the result consistent with the REST API


iLLiCiTiT (approved):

iLLiCiTiT:
If it doesn't break anything on frontend, it should be good for most of ayon-api usages.

PR #495, opened 2025-01-22
Events: Horizontal scaling support in EventStream.subscribe

martastain:
This pull request introduces several key improvements to the handling of event subscribers and the organization of hooks in the EventStream class.

####  Single Instance Handling

When an event is created and a subscriber is attached to the topic, the function is triggered only on the current server instance.

Use Case: Situations where the event should be processed once, e.g., "Entity status change sends a notification."

#### All Instances Handling:
In scaled deployments of Ayon (multiple server instances), some events need to trigger functions on all instances.

Use Case: "Addon settings change requires all addon instances to reload their cache."

### Implementation

Added a new parameter, `all_nodes: bool`, to the `EventStream.subscribe method` (Default: False)
When set to True, the subscribed function will be triggered on all server instances.

Hooks are now split into two groups:
- `local_hooks`: Implementation remains unchanged.
- `global_hooks`: Implementation added in ayon_server.api.messaging.

### Testing Notes

Use multiple replicas in a Docker Compose environment.

```yaml
server:
    image: ynput/ayon:dev
    restart: unless-stopped
    deploy:
      replicas: 3
    ports: 
      -"5000-5002:5000"
```

Subscribe to an event topic from an addon using:

```python
EventStream.subscribe(topic_name, handler, all_nodes=True)
```

Verify correct behavior for both single-instance and all-instance event handling.

Keep in mind Ayon server is still not fully horizontally scalable, so other functionality may break when the server runs with multiple replicas.

### DIFF
--- a/ayon_server/events/eventstream.py
+++ b/ayon_server/events/eventstream.py
@@ -23,36 +23,18 @@ class EventStream:
         cls, topic: str, handler: HandlerType, all_nodes: bool = False
     ) -> str:
         token = create_id()
-        if all_nodes:
-            if topic not in cls.global_hooks:
-                cls.global_hooks[topic] = {}
-            cls.global_hooks[topic][token] = handler
-            return token
-        else:
-            if topic not in cls.local_hooks:
-                cls.local_hooks[topic] = {}
-            cls.local_hooks[topic][token] = handler
-            return token
+        hooks = cls.global_hooks if all_nodes else cls.local_hooks
+        topic_hooks = hooks.setdefault(topic, {})
+        topic_hooks[token] = handler
+        return token
 
     @classmethod
     def unsubscribe(cls, token: str) -> None:
-        # Local hooks
-        topics_to_remove = []
-        for topic in cls.local_hooks:
-            cls.local_hooks[topic].pop(token, None)
-            if not cls.local_hooks[topic]:
-                topics_to_remove.append(topic)
-        for topic in topics_to_remove:
-            cls.local_hooks.pop(topic)
-
-        # Global hooks
-        topics_to_remove = []
-        for topic in cls.global_hooks:
-            cls.global_hooks[topic].pop(token, None)
-            if not cls.global_hooks[topic]:
-                topics_to_remove.append(topic)
-        for topic in topics_to_remove:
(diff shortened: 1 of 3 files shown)


ayon_server/events/eventstream.py
```diff
@@ -15,25 +15,44 @@
 
 class EventStream:
     model: Type[EventModel] = EventModel
-    hooks: dict[str, dict[str, HandlerType]] = {}
+    local_hooks: dict[str, dict[str, HandlerType]] = {}
+    global_hooks: dict[str, dict[str, HandlerType]] = {}
```
iLLiCiTiT, line 19: With this you can get rid of multiple `if topic not in *hooks*` conditions and avoid possible missing keys issues.
```suggestion
    local_hooks: dict[str, dict[str, HandlerType]] = collections.defaultdict(dict)
    global_hooks: dict[str, dict[str, HandlerType]] = collections.defaultdict(dict)
```

ayon_server/events/eventstream.py
```diff
@@ -15,25 +15,44 @@
...
+        if all_nodes:
+            if topic not in cls.global_hooks:
+                cls.global_hooks[topic] = {}
+            cls.global_hooks[topic][token] = handler
+            return token
+        else:
+            if topic not in cls.local_hooks:
+                cls.local_hooks[topic] = {}
+            cls.local_hooks[topic][token] = handler
+            return token
```
iLLiCiTiT, line 35:
```suggestion
        hooks = cls.global_hooks if all_nodes else cls.local_hooks
        topic_hooks = hooks.setdefault(topic, {})
        topic_hooks[token] = handler
        return token
```

ayon_server/events/eventstream.py
```diff
@@ -15,25 +15,44 @@
...
 
     @classmethod
     def unsubscribe(cls, token: str) -> None:
+        # Local hooks
```
iLLiCiTiT, line 39: Works, but can be improved (multiple different ways):
```python
        for hooks in (cls.global_hooks, cls.local_hooks):
            for topic, mapping in tuple(hooks.items()):
                mapping.pop(token, None)
                if not mapping:
                    hooks.pop(topic)
```

PR #506, opened 2025-02-04
Settings: Unify saving settings

martastain:


martastain:
> I am not sure if I test this PR correctly. Everything is logged correctly if I use settings UI. However, nothing is logged when lower-lever editor is used.

That's expected - we cannot use this function for low-level edits as it requires valid data, while low-level editor is especially useful when data is inconsistent/broken. But maybe we could at least log the fact someone used the low level editor.

64qam (approved):

PR #507, opened 2025-02-05
REST: tasksFolders endpoint

martastain:
This new endpoint returns a list of folder ids containing tasks matching the given query. This is also a prototype of the proposed query structure that may be adopted for more entity types.

The query supports multiple nested conditions. The `filter` object accepts two parameters:

- **`operator`**: Either `"and"` or `"or"` (defaults to `"and"` if not provided).
- **`conditions`**: A list of conditions, which can either be:
  - Another `filter` object (allowing for nested queries).
  - A condition object containing `key`, `operator`, and `value`.

### Condition Format

Each condition consists of:
- **`key`**: The name of the field (e.g., `status`, `tags`, `assignees`, etc.).
- **`operator`**: A comparison method.
- **`value`**: The value to compare against (if applicable).

### Supported Operators

| Operator  | Description |
|-----------|------------|
| `eq`      | Stored value must match the queried value. |
| `lt`      | Match records with value lower than the requested value  |
| `gt`      | Match records that are greater than the requested value. |
| `lte`     | Match records with value lower or equal the requested value |
| `gte`     | Match records with value greater or equal the requested value |
| `ne`      | Queried value must not match the stored value. |
| `isnull`  | Stored value must be `null` (`value` field is ignored). |
| `notnull` | Stored value must not be `null` (`value` field is ignored). |
| `in`      | Stored value must match at least one item in the array provided in `value`. |
| `notin`   | Stored value must not be present in the array provided in `value`. |
| `contains`| Stored value (array) must contain the provided `value`. |
| `excludes`| Stored value (array) must not contain the provided `value`. |
| `any`     | Both queried and stored values are arrays, and they must have at least one common element. |
| `like`    | Performs an SQL-like partial string match. |

This structure provides flexibility in querying tasks while maintaining consistency across different entity types.

`[POST] /api/projects/myproject/tasksFolders`

```json
{
    "filter": {
        "operator": "and",
        "conditions": [
            {
                "key": "tags",
                "operator": "any",
                "value": ["fluffy", "cute"],
            },
            {
                "operator": "or",
                "conditions": [
                    {
                        "key": "status",
                        "operator": "eq",
                        "value": "In progress"
                    },
                    {
                        "key": "status",
                        "operator": "eq",
                        "value": "Pending review"
                    },

                ]
            }
        ]
    }
}
```

Translates to SQL condition:

```sql
SELECT DISTINCT(folder_id) FROM project_myproject.tags 
WHERE
    tags::text[] && array['fluffy'] 
AND (status = 'In progress' OR status = 'Pending review'))
```

So the result will contain a list of ids of folders, that contain tasks with tag "fluffy" or "cute" and simultaneously have status `In progress` or `Pending review`

### DIFF
--- a/api/events/enroll.py
+++ b/api/events/enroll.py
@@ -14,7 +14,7 @@ from ayon_server.exceptions import (
 )
 from ayon_server.lib.postgres import Postgres
 from ayon_server.lib.redis import Redis
-from ayon_server.sqlfilter import Filter
+from ayon_server.sqlfilter import QueryFilter
 from ayon_server.types import TOPIC_REGEX, Field, OPModel
 from ayon_server.utils import hash_data
 
@@ -59,7 +59,7 @@ class EnrollRequestModel(OPModel):
         description="Ensure events are processed in sequential order",
         example=True,
     )
-    filter: Filter | None = Field(
+    filter: QueryFilter | None = Field(
         None, title="Filter", description="Filter source events"
     )
     max_retries: int = Field(3, title="Max retries", example=3)
--- a/api/events/operations.py
+++ b/api/events/operations.py
@@ -3,7 +3,7 @@ from typing import Literal
 from ayon_server.api.dependencies import CurrentUser
 from ayon_server.exceptions import ForbiddenException, NotImplementedException
 from ayon_server.lib.postgres import Postgres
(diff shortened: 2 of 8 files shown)


martastain:
added option to add `filter: "JSON_ENCODED_QUERY"` to graphql tasks resolver

PR #520, opened 2025-02-20
New logging system and utils refactoring

martastain:
This PR removes the dependency on the `nxtools` library and reimplements all necessary functions within the `ayon_server` module.  

The most significant change is the transition to `loguru` for logging. Log levels are now configurable for both console output and the event stream. By setting the `AYON_LOG_MODE=json` environment variable, console logging can be switched to structured JSON format, providing additional context. All log messages in the event stream now include metadata such as the originating file, function, and line number.  

Additionally, logs can be collected and visualized using the **Promtail + Loki + Grafana** stack, enabling better insight into log relationships. Further improvements in this area are planned for future updates.  

Fake `nxtools` module was added to the server and all used functions are aliased, so this PR is fully backwards compatible.
It is recommended to adapt `ayon_server.logging` and `ayon_server.utils` imports in addons (imports have been already updated in the server itself)

image
image
image


ayon_server/logging.py
```diff
@@ -0,0 +1,66 @@
...
+        level = record["level"].name
+        message = record["message"]
+        module = record["name"]
+        formatted = f"{level:<8} | {module:<25} | {message}"
+        write_stderr(formatted)
+        if tb := record.get("traceback"):
+            write_stderr(indent(str(tb)))
+
+
+def get_logger():
```
iLLiCiTiT, line 46: Is this function something what should be used in addons for example (or anywhere else)?

martastain: No. when addons import `from nxtools import logging` or `from ayon_server.logging import logging` they get this instance, and loguru handles the contextualization.

ayon_server/logging.py
```diff
@@ -0,0 +1,66 @@
...
+    """Log the current exception traceback."""
+    tb = traceback.format_exc()
+    logging.error(message, traceback=tb, **kwargs)
+
+
+def critical_error(message="Critical Error!", **kwargs):
+    """Log a critical error message and exit the program."""
+    logging.critical(message, **kwargs)
+    logging.error("Exiting program.")
+    sys.exit(1)
```
iLLiCiTiT, line 66:
```suggestion
logger = get_logger()


def log_traceback(message="Exception!", **kwargs):
    """Log the current exception traceback."""
    tb = traceback.format_exc()
    logger.error(message, traceback=tb, **kwargs)


def critical_error(message="Critical Error!", **kwargs):
    """Log a critical error message and exit the program."""
    logger.critical(message, **kwargs)
    logger.error("Exiting program.")
    sys.exit(1)


def debug(*args, **kwargs):
     logger.debug(*args, **kwargs)
...
```

ayon_server/models/rest_model.py
```diff
@@ -1,10 +1,10 @@
 from typing import Any
 
-from nxtools import logging
 from pydantic import BaseModel
 from pydantic.fields import FieldInfo, Undefined
 from pydantic.typing import NoArgAnyCallable
 
+from ayon_server.logging import logging
```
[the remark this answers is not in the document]
iLLiCiTiT: from ayon_server import logging

antirotor:
Maybe stupid remark about f-strings in the logger - isn't it the same issue as with the native python logger that it is hindering performance because the strings are formatted even when the log might not be emitted?

PR #521, opened 2025-02-21
Backend support for project bundles

martastain:
This pull request introduces several changes to the `api` and `ayon_server` modules to support project-specific addon versions and project bundles. The most important changes include adding new fields to models, updating bundle handling logic, and introducing new endpoints for project bundles.


iLLiCiTiT:
I can create dev project bundle, that probably should not be allowed.

iLLiCiTiT:
For some reason I cannot create project bundle now 
image

martastain:
> Ok, it happens when I duplicate project bundle, UI shows current studio bundle addon versions, but the value in backend is None. If I clone studio bundle with filled version, or create new bundle, or fill the versions and then switch to project, it works.

The UI is really just a proof-of-work. @Innders will need to look at it and polish it :-I

antirotor (approved):
I've tested bundles creation, deletion, usage of specific bundles, listing bundles - all works well.

kalisp (approved):
Looks good, played with bundles for a while, updated settings, used dev/prod/staging in my launcher, values were picked correctly from Settings.

MustafaJafar (approved):
I was able to find the new changes in the api. (didn't test them, my main review about backward compitablity)
image

tested launcher in dev and production modes.
testing changing few settings in dev and prod bundles.
everything seems to work as usual.

also, I was able to create bundles and copy settings.

PR #526, opened 2025-03-03
Events: emit granular events when project is updated

martastain:
This pull request introduces a new function to build project change events and integrates it into the existing project update and anatomy setting functionalities. The main goal of these changes is to enhance the event dispatching mechanism when project data is modified.

The following new events are introduced. All of them support audit trail (`oldValue` and `newValue` stored in the event payload)

- [x] Emit `entity.project.attrib_changed` event when project attribute is changed using patch request or setting its anatomy
- [x] Emit `entity.project.config_changed`
- [x] Emit `entity.project.active_changed`
- [x] Emit `entity.project.data_changed`

The original `entity.project.changed` is still being emited to maintain the backwards compatibility (it does not contain the audit trail)

### DIFF
--- a/api/projects/anatomy.py
+++ b/api/projects/anatomy.py
@@ -2,6 +2,7 @@ from ayon_server.api.dependencies import CurrentUser, ProjectName, Sender, Sende
 from ayon_server.api.responses import EmptyResponse
 from ayon_server.entities import ProjectEntity
 from ayon_server.events import EventStream
+from ayon_server.events.patch import build_project_change_events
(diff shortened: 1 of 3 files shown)


iLLiCiTiT (approved):
It still raises `entity.project.changed` but I guess we should keep it for backwards compatibility.

martastain:
> It still raises `entity.project.changed` but I guess we should keep it for backwards compatibility.

Yep. That was the idea. we could eventually hide it from the viewer and then deprecate it.

PR #530, opened 2025-03-04
Operations: add error code to operation result

martastain:
Adds errorCode field to operation result to allow granular error handling


BigRoy (approved):
image

Seems to work.
Works for both 'duplicate' task names under a folder, and duplicate folder names under a parent.

Deletion also still works with its original correct message:
image

PR #562, opened 2025-04-02
New version status respects filtering by Scope

kalisp:
## Description of changes

Each Status could be set specific entity types ('Folder', 'Task', 'Version'). Querying of default status for new entity ignored if said entity_type was unset in `Scope`.

### Technical details

I am not completely sold on second commit where I am removing old filtering on `entity_types` and `entity_subtypes`. It seems obsolete and replaced by `Scope` to me.
Another question would be if `entity_subtype` shouldn't be removed altogether (but question for different issue).

### Testing notes
- remove 'Version' from first (or multiple) Status 
image
- newly created Version should use next available Status (by default it was always `Not ready`)
image


martastain (approved):
Agreed it most likely used a deprecated fields.

MilaKudr (approved):

PR #568, opened 2025-04-07
Streamlined array comparison in SQLFilter

martastain:
# Comparison Operator Documentation

This table outlines the available comparison operators used in the query filter system, including their meaning, SQL translation, and additional notes.
| Operator    | Description                                      | SQL Expression                          |
|-------------|--------------------------------------------------|-----------------------------------------|
| `eq`          | Equal to                                         | column = value                          |
| `ne`          | Not equal to                                     | column != value                         |
| `lt`          | Less than                                        | column < value                          |
| `gt`          | Greater than                                     | column > value                          |
| `lte`         | Less than or equal to                            | column <= value                         |
| `gte`         | Greater than or equal to                         | column >= value                         |
| `like`        | String comparison using LIKE                     | column LIKE value                       |
| `isnull`      | Field is NULL                                    | column IS NULL                          |
| `notnull`     | Field is NOT NULL                                | column IS NOT NULL                      |
| `in`          | Field value is in the given list                 | column = ANY(array[value1, value2])     |
| `notin`       | Field value is not in the given list             | NOT column = ANY(array[value1, value2]) |
| `includes`    | Field contains the scalar value                  | value = ANY(column)                     |
| `excludes`    | Field does not contain the scalar value          | NOT (value = ANY(column))               |
| `includesall` | Field array includes all provided values         | column @> array[value1, value2]         |
| `excludesall` | Field array does not include all provided values | NOT (column @> array[value1, value2])   |
| `includesany` | Field array includes any of the provided values  | column && array[value1, value2]         |
| `excludesany` | Field array does not contain any of the provided values  | NOT (column && array[value1, value2])   |

### DIFF
--- a/ayon_server/sqlfilter.py
+++ b/ayon_server/sqlfilter.py
@@ -278,7 +278,7 @@ def build_condition(c: QueryCondition, **kwargs) -> str:
     elif operator == "excludes":
         return f"NOT ({safe_value} = ANY({column}))"
     elif operator == "like":
-        return f"{column} LIKE {safe_value}"
+        return f"{column} ILIKE {safe_value}"
     else:
         raise ValueError(f"Unsupported operator: {operator}")
 
--- a/ayon_server/sqlfilter.py
+++ b/ayon_server/sqlfilter.py
@@ -20,6 +20,7 @@ ValueType = (
 
 OperatorType = Literal[
     "eq",
+    "like",
     "lt",
     "gt",
     "lte",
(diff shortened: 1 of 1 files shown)


martastain:
slightly changed behavior:

`excludesany`: Reject if it contains at least one of the values.
`excludes`: Match only if none of the values are present — that is, there’s no overlap at all.

Innders (changes requested):
```
{
    "conditions": [
        {
            "key": "assignees",
            "value": [
                "becky",
                "luke"
            ],
            "operator": "excludesany"
        }
    ],
    "operator": "and"
}
```

Is `exludesany` working here? I would have thought it would filter out any task that has assignee becky or luke.

image

Same for this one:

```
{
    "conditions": [
        {
            "key": "assignees",
            "value": [
                "becky",
                "luke"
            ],
            "operator": "excludes"
        }
    ],
    "operator": "and"
}
```

image

(I definitely have the latest changes because the generated files have the new `excludesany` type.

PR #569, opened 2025-04-07
Chore: Templates enum

iLLiCiTiT:
## Description of changes
Small enhancements for templates enum for settings.

### Technical details
Removed unnecessary `project_name` argument from enum function `anatomy_template_items_enum`. Whatever would be passed to the function would not be used anyways, so why to bother developer by having argument that does nothing?

Removed outdated information from docstring.

Avoid creation of <dict.Key> -> <list[str]> -> iterable[str] just to extend list.

### DIFF
--- a/ayon_server/settings/enum.py
+++ b/ayon_server/settings/enum.py
@@ -147,8 +147,6 @@ def anatomy_template_items_enum(
 
     Wrapper for actual function as Settings require callable.
 
-    If no category is required it must be used as:
-        `enum_resolver=anatomy_template_items_enum()`
     Args:
         project_name: str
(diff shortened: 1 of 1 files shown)


kalisp:
I think `project_name` had an impact, it was differentiating between Studio and Project Settings, which could contain different templates. (I think.)

iLLiCiTiT:
> I think project_name had an impact, it was differentiating between Studio and Project Settings, which could contain different templates. (I think.)

Function `anatomy_template_items_enum` does not need to expect it, you always must call the function and it returns new partial function. And the partial function which is returned, has to expect `project_name`, which does.

martastain (approved):
tested. works

PR #577, opened 2025-04-23
Use timestamptz casting when sorting timestamps

martastain:
This pull request introduces minor logging enhancements and a new feature to handle ISO timestamp values in cursor decoding for GraphQL resolvers. 

Added a regular expression check to identify ISO timestamp values in the `decode_cursor` function. If a match is found, the value is converted to a PostgreSQL `timestamptz` format for proper handling.


Innders (approved):
Works great thanks!

PR #591, opened 2025-05-07
Entity list items filtering

martastain:
This pull request introduces significant enhancements to the filtering and sorting capabilities of the `entity_list_items` query in the GraphQL API, along with related changes to improve maintainability and functionality. The most important updates include adding a new `filter` parameter to support advanced filtering, refactoring column naming conventions for consistency, and implementing unified attributes for sorting and filtering. 

### Unified attributes

For filtering and sorting `attrib` field unifies entity attributes with list items attributes:

Assuming we're loading a list of tasks:

```graphql
items(sortBy: "attrib.priority"){
  ...
}
```

Sorting will use the following fields to determine the actual value to sort by in this order:

- list item attribute (if the priority is set on the item level)
- task attribute
- folder attribute
- parent folder or project attribute

The same logic is applied for filtering

```graphql
items(filter:"{\"conditions\":[{\"key\": \"attrib.priority\", \"value\":\"high\"}]}){
  ...
}
```

### Entity fields

When filtering or sorting by entity fiels (such as `id`, `status`, `name`, `label`, etc. ). It is necessary to prefix the field with `entity_`. 
The algorithm normalizes snake/camel case variants of the fields, so the following variants will produce the same result.

```graphql
items(sortBy: "entity_created_at"){
  ...
}

items(sortBy: "entity_createdAt"){
  ...
}
```

It is possible to filter by all non-dynamic fields of an entity. That means you cannot use:

- `entity_parents` for folder lists
- `entity_name` for version lists

and similar (these fields are not stored in the database, but generated on the fly)

### Item fields

It is possible to sort by the following item fields, by default graphql sorts by `position`:

- `position`
- `label`
- `createdAt`
- `updatedAt`
- `createdBy`
- `updatedBy`
- `folderPath`

Again, both camelCase and snake_case variants are supported. This is unique to list_items and will be gradually adopted in other resolvers

Additionally, filtering accepts all fields as the sorter, plus the following ones:

- `id`
- `entityId`
- `entityListId`
- `data`
- `tags`

### DIFF
--- a/ayon_server/api/logging.py
+++ b/ayon_server/api/logging.py
@@ -110,6 +110,7 @@ def handle_undhandled_exception(request: Request, exc: Exception) -> JSONRespons
 
     path_prefix = f"{os.getcwd()}/"
     formatted = "".join(traceback.format_exception_only(type(exc), exc)).strip()
+    formatted = formatted.replace("{", "{{").replace("}", "}}")
     tb = traceback.extract_tb(exc.__traceback__)
     traceback_msg = ""
     for frame in tb[-20:]:
--- a/ayon_server/graphql/__init__.py
+++ b/ayon_server/graphql/__init__.py
@@ -207,7 +207,7 @@ class AyonSchema(strawberry.Schema):
                 else:
                     path = ""
 
-                message = error.message
+                message = error.message.replace("{", "{{").replace("}", "}}")
                 logger.error(
                     f"[GRAPHQL] Error resolving {path} (line {line_no}): {message}",
                     module=fname,
--- a/ayon_server/graphql/nodes/entity_list.py
+++ b/ayon_server/graphql/nodes/entity_list.py
@@ -196,6 +196,7 @@ class EntityListNode:
         before: str | None = None,
         sort_by: str | None = None,
         accessible_only: bool = False,
+        filter: str | None = None,
     ) -> EntityListItemsConnection:
         if first is None and last is None:
             first = 200
@@ -209,6 +210,7 @@ class EntityListNode:
             last=last,
             before=before,
             sort_by=sort_by,
+            filter=filter,
             accessible_only=accessible_only,
         )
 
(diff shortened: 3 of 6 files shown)


Innders:
Is it possible to filter entities by parent type.
Filter: 
`versions` by `productType`, `taskType`, `folderType`
`products` by `productType`, `taskType`, `folderType`
`tasks` by `taskType`, `folderType`
`folders` by `folderType`

Innders (changes requested):
boolean attrib filters causes errors.

```gql

    query GetTasksByParent($projectName: String!, $parentIds: [String!]!, $filter: String, $search: String) {
  project(name: $projectName) {
    name
    tasks(folderIds: $parentIds, filter: $filter, search: $search) {
      edges {
        node {
          ...TaskPropsFragment
        }
      }
    }
  }
}
    
    fragment TaskPropsFragment on TaskNode {
  id
  folderId
  label
  name
  ownAttrib
  status
  tags
  taskType
  updatedAt
  active
  assignees
  allAttrib
  folder {
    path
  }
}
```

``json
{
  "projectName": "demo_Commercial",
  "parentIds": [
    "72712dda268d11f09ccd0242ac130004"
  ],
  "filter": "{\"conditions\":[{\"key\":\"hard\",\"value\":false,\"operator\":\"eq\"}],\"operator\":\"and\"}"
}
```

```json
{
  "data": null,
  "errors": [
    {
      "message": "7 validation errors for QueryFilter\nconditions -> 0 -> value\n  str type expected (type=type_error.str)\nconditions -> 0 -> value\n  value is not a valid integer (type=type_error.integer)\nconditions -> 0 -> value\n  value is not a valid float (type=type_error.float)\nconditions -> 0 -> value\n  value is not a valid list (type=type_error.list)\nconditions -> 0 -> value\n  value is not a valid list (type=type_error.list)\nconditions -> 0 -> value\n  value is not a valid list (type=type_error.list)\nconditions -> 0 -> operator\n  unexpected value; permitted: 'and', 'or' (type=value_error.const; given=eq; permitted=('and', 'or'))",
      "locations": [
        {
          "line": 5,
          "column": 5
        }
      ],
      "path": [
        "project",
        "tasks"
      ]
    }
  ]
}
```

Innders (approved):

PR #599, opened 2025-05-15
List actions

martastain:
This pull request updates the `ActionContext` class in `ayon_server/actions/context.py` to include support for a new entity type, `"list"`. It also adjusts the logic in the `get_entities` method to account for this addition.

### action.entity_subtypes

- always includes the contained entity type (version, folder, etc.)
- optionally includes the list type: "version:review-session"

### context.entity_subtypes

always includes both parts: "version:review-session"

### Matching Rules

An action_subtype matches a context_subtype if:

- The contained type matches (version == version)
- If action_subtype includes a list type, it must also match (review-session)
- If action_subtype doesn't include a list type,  it's considered a wildcard (matches all list types  of that entity type)

### Example

manifest subtype | context subtype | Match?
-- | -- | --
`version` | `version:review-session` | ✅
`version:review-session` | `version:review-session` | ✅
`version:some-other` | `version:review-session` | ❌
`folder` | `version:review-session` | ❌

### DIFF
--- a/api/actions/listing.py
+++ b/api/actions/listing.py
@@ -93,6 +93,53 @@ async def get_relevant_addons(user: UserEntity) -> tuple[str, list[BaseServerAdd
     )
 
 
+def match_list_subtypes(
+    action_subtypes: list[str],
+    context_subtypes: list[str],
+) -> bool:
+    """List matching
+
+    ## action subtype (defined by manifest)
+
+    - always includes the contained entity type (version, folder, etc.)
+    - optionally includes the list type: "version:review-session"
+
+    ## context subtypes (provided by the client)
+
+    always includes both parts: "version:review-session"
+
+    ## Matching Rules
+
+    An action_subtype matches a context_subtype if:
+
+    - The contained type matches (version == version)
+
+    - And if action_subtype includes a list type,
+      it must also match (review-session)
+
+    - If action_subtype doesn't include a list type,
+      it's considered a wildcard (matches all list types
+      of that entity type)
+    """
+    for context_subtype in context_subtypes:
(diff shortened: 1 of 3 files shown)


api/actions/listing.py
```diff
@@ -93,6 +93,53 @@ async def get_relevant_addons(user: UserEntity) -> tuple[str, list[BaseServerAdd
...
+    - The contained type matches (version == version)
+
+    - And if action_subtype includes a list type,
+      it must also match (review-session)
+
+    - If action_subtype doesn't include a list type,
+      it's considered a wildcard (matches all list types
+      of that entity type)
+    """
+    for context_subtype in context_subtypes:
```
iLLiCiTiT, line 124: Shouldn't this be more like this?
```suggestion
    if not action_subtypes or None in action_subtypes:
        return True

    for context_subtype in context_subtypes:
```

api/actions/listing.py
```diff
@@ -117,8 +164,16 @@ async def evaluate_simple_action(
...
+            if action.entity_type == "list":
+                return match_list_subtypes(
+                    action.entity_subtypes,
+                    context.entity_subtypes,
+                )
+
+            else:
+                # Normal project level entities. Just a simple subtype match
+                if not set(action.entity_subtypes) & set(context.entity_subtypes):
+                    return False
```
iLLiCiTiT, line 176: No need for else
```suggestion
            # Normal project level entities. Just a simple subtype match
            if not set(action.entity_subtypes) & set(context.entity_subtypes):
                return False
```

iLLiCiTiT (approved):

PR #616, opened 2025-05-28
Support 'syntax' argument for settings field

martastain:
This pull request introduces a new `syntax` parameter to the `SettingsField` function in the `ayon_server/settings/settings_field.py` file. The changes ensure that the `syntax` parameter is only applicable when the `widget` type is set to "textarea" and logs a debug message if used incorrectly.

### DIFF
--- a/ayon_server/settings/settings_field.py
+++ b/ayon_server/settings/settings_field.py
@@ -43,6 +43,7 @@ def SettingsField(
     required_items: list[str] | None = None,
     section: str | None = None,
     widget: str | None = None,
+    syntax: str | None = None,
     layout: str | None = None,
     tags: list[str] | None = None,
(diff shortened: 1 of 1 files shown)


martastain:
Yeah. There should be a strict type, but i'm not sure yet, what can be supported in the frontend. I'll create a follow up PR after more tests.

As for **kwargs - it's there for backwards compatibility with very old addons - it logs a message that argument is not supported when there's something catched in there.

BigRoy:
There's a bit of a downside to the syntax highlight colors.
I wonder if e.g. this is nice:
image

Or whether it feels too much like the "blue" override color.
Tagging @mkolar @Innders for visibility.

If it's nice I can make some PRs to add it to common settings for `ayon-core` and a few others.

PR #637, opened 2025-06-23
REST:  hasReviewables flag in list_folders endpoint

martastain:
This pull request introduces changes to support the concept of "reviewables" in the hierarchy cache and related functionality. Key updates include adding the `has_reviewables` attribute to folder entities, modifying the hierarchy cache logic to account for reviewables, and simplifying the `rebuild_hierarchy_cache` function by removing the `transaction` parameter. Below is a breakdown of the most important changes.

## Testing notes

It may be necessary to do at least some write operation in the project to invalidate the current hierarchy cache

### DIFF
--- a/api/folders/list_folders.py
+++ b/api/folders/list_folders.py
@@ -76,6 +76,7 @@ class FolderListItem(OPModel):
     folder_type: str
     has_tasks: bool = False
     has_children: bool = False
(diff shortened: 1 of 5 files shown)


Innders (changes requested):
Works great thanks!

image

PR #639, opened 2025-06-23
Simplified transactions

martastain:
This pull request refactors database interaction across several modules by standardizing the use of the `Postgres` helper class and removing direct references to `conn`. It also enforces transactional integrity by adding assertions to ensure certain functions are executed within a transaction. Additionally, redundant code has been removed or simplified for better readability and maintainability.

## testing notes

Consider everything may be broken.

- operations 
- bundles
- settings migration
- activities
- ...

Concurrency should improve (especially concurrent entity updates).

When multiple requests tries to update one entity at the VERY same time, one of the request will fail after 3 retries and returns 503 service unavailable status signalling that update should be attempted later (it should be an edgy edge case)


api/bundles/actions.py
```diff
@@ -2,16 +2,21 @@
...
-async def promote_bundle(bundle: BundleModel, user: UserEntity, conn):
+async def promote_bundle(bundle: BundleModel, user: UserEntity):
     """Promote a bundle to production.
 
     That includes copying staging settings to production.
     """
 
+    assert (
+        await Postgres.is_in_transaction()
+    ), "promote_bundle function must be called within a transaction"
```
BigRoy, line 18: I wonder whether stuff like this warrants being a function decorator? So you just have `@require_transaction` on the function or something.

ayon_server/entities/folder.py
```diff
@@ -67,160 +68,146 @@ async def load(
...
-                    WHERE path = $1
-                    OR path LIKE $1 || '/%'
-                ) RETURNING name
-                """,
-                self.path.lstrip("/"),
-            )
+    async def delete(self, *args, auto_commit: bool = True, **kwargs) -> bool:
+        async with Postgres.transaction():
+            if kwargs.get("force", False):
+                logger.info(f"Force deleting folder and all its children. {self.path}")
```
BigRoy, line 190: I assume `:` may make more sense on this logged message?
```suggestion
                logger.info(f"Force deleting folder and all its children: {self.path}")
```

ayon_server/operations/project_level/__init__.py
```diff
@@ -133,17 +130,26 @@ async def _process_operations(
...
-            result.append(response)
-            if entity.entity_type not in [e.entity_type for e in to_commit]:
-                to_commit.append(entity)
+            # This is a neat trick. transaction() will try
+            # to reuse the current transaction if it exists,
+            # but create a new one if it doesn't.
+
+            # This way, every operation is ensured to run in a transaction,
+            # but main process may wrap everything in a single one
+            # to commit all operations at once.
```
BigRoy, line 139: I'm not sure what the size of the of the 'transaction' may become at this point, but if it may grow large with more and more stacking into a transaction. Then it may be good to be aware of potential downsides due to it, e.g. [locking](https://dba.stackexchange.com/a/273905). Rows touched by the transaction will remain locked for the remainder duration of the transaction. This may make sense, but beware that (according to that page?) even reading the row may be blocked _during_ that transaction?

In short, transactions may limit the possibility of concurrency.

_I know nothing of databases NOR of this part of the code, so I'm likely saying very stupid stuff here._

ayon_server/operations/project_level/__init__.py
```diff
@@ -415,29 +425,44 @@ async def _process(
...
-                async with Postgres.acquire() as conn, conn.transaction():
-                    events, response = await _process_operations(
-                        self.project_name,
-                        self.operations,
-                        user_map=self.user_entities_map,
-                        transaction=conn,
-                        raise_on_error=raise_on_error,
-                    )
-
+            for _ in range(3):
```
BigRoy, line 428: Should this be some global retry constant? or?

api/bundles/actions.py
martastain: Good idea. I'll put that on the todo list for the next iteration

ayon_server/operations/project_level/__init__.py
martastain: Okay. We may figuer out later, what is the same number. During my stress tests, the first retry was usually successful.

MilaKudr (approved):
I mainly tested the sync from SG to Ayon, as this was the main issue, and all changes are reflected in Ayon correctly. The number of connections to the postgres database is around 10 most of the time. I think update requests are processed more quickly.

PR #649, opened 2025-07-10
Basic OAuth and JWT support

antirotor:
# OAuth Provider Implementation for AYON Server

This implementation provides a basicOAuth 2.0 authorization server for AYON Server, allowing third-party applications to securely access user data through standardized OAuth processes. It is using [oauthlib](https://oauthlib.readthedocs.io/en/latest/)

### Authorization Endpoints

- `GET /api/oauth/authorize` - OAuth authorization endpoint
- `POST /api/oauth/consent` - User consent handling
- `POST /api/oauth/token` - Token exchange endpoint
- `POST /api/oauth/introspect` - Token introspection endpoint
- `GET /api/oauth/userinfo` - User information endpoint

### JWT Token Endpoints

- `POST /api/oauth/jwt` - Generate JWT tokens (authenticated users)
- `POST /api/oauth/jwt/exchange` - Exchange OAuth tokens for JWT tokens
- `GET /api/oauth/validate` - Validate JWT tokens
- `GET /.well-known/jwks.json` - JSON Web Key Set

### Client Management (Admin Only)

- `GET /api/oauth/clients` - List OAuth clients
- `POST /api/oauth/clients` - Create OAuth client
- `GET /api/oauth/clients/{client_id}` - Get OAuth client
- `DELETE /api/oauth/clients/{client_id}` - Delete OAuth client

### Discovery

- `GET /.well-known/openid_configuration` - OpenID Connect Discovery

## Notes

* Clients are presisted in DB but tokens are Redis-based. 
* This PR doesn't handle scopes and all the proper token validations and processing (like client can be determined from the provided token, etc. - some logic relies on currently logged user). 
* Consent page (the page where the user is redirected to express consent about using his information to log on connecting client) is missing and needs to be implemented in the frontend.
* Tests will be added to test repo


antirotor:
So I've tested it and it sort of works but there are several weak points/questions to be answered

### access token vs. oauth
once you get oauth access token, you shouldn't need ayon access token but you still do. Maybe we should support `Authentication: Bearer ...` as well as `x-api-key`

### JWT issuer
I wasn't able to figure out how to get the server name without access to Request object. Now, issuer is stupidly hardcoded and not validated.

### Keys
Now we use `HS256` symmetrical key, but we should use `RS256` asymmetric ones

### JWT endpoint
Is very useful but right now it is using some hardcoded values. Maybe this should be turned into helper method to be used in individual addons for better context?

antirotor:
you can use this little scripts for testing btw: https://gist.github.com/antirotor/7f5720a5ba168edac2c6c31825f46a2b

PR #659, opened 2025-07-23
Actions: List actions endpoint allows to specify settings variant

iLLiCiTiT:
## PR Checklist
-   [ ] `/api/actions/list` endpoint allows to explicitly define settings variant.

## Description of changes
It was not possible to receive actions for explicitly defined settings variant which is a big issue for pipeline purposes. There was no way how to get `staging` actions, or actions for specific dev bundle.

### Technical details
The endpoint is backwards compatible and still works as before until the `variant` query is passed to the endpoint.

Resolves https://github.com/ynput/ayon-backend/issues/651


martastain (approved):
Code looks good, doesn't seem to break anything.

PR #684, opened 2025-08-18
Set version watchers before version.publish activity is created

martastain:
This pull request refactors the way the `version.publish` activity is created in the `handle_version_created` method of `ayon_server/activities/event_hook.py`. The main change is that the activity creation has been moved to occur after watcher-related logic, rather than before.

This ensures task watchers are notified about a new version being created.


Innders (approved):
Looks good to me!

Image

PR #712, opened 2025-09-15
Provide current user's access level in entity list objects

### LINKED ISSUE #709: Lists: AccessLevel for current user
Lists have access control and we need to know the resolved access level for the current user.

This is required so that we know on the client if the user can view/edit/delete a specific list.

martastain:
This pull request enhances access control for entity lists in the GraphQL API by introducing a new `access_level` field on EntityList node and supporting infrastructure.


Innders (approved):
Tested and works

PR #736, opened 2025-09-29
Add overview row height to view data model

martastain:
This pull request introduces a minor update to the `OverviewSettings` model in `ayon_server/views/models.py`. The change adds an optional `row_height` field to allow configuration of row height in the overview settings.


filipvnencak (changes requested):
Thanks, but add also row_height prop to `/api/views/lists`

filipvnencak (approved):
Working nice

PR #738, opened 2025-09-30
Configurable entity naming

martastain:


Innders (commented):
Both fields have enum so we know the types of both are specific strings.

The generated types are just strings

```tsx
export type EntityNaming = {
  /** How to capitalize the entity names */
  capitalization?: string
  /** Character to separate different parts of the name */
  separator?: string
}
```

Is it possible to have the types?

```tsx
export type EntityNaming = {
  /** How to capitalize the entity names */
  capitalization?: 'lower' | 'upper' | 'keep' | 'capitalize'
  /** Character to separate different parts of the name */
  separator?: '' | '_' | '-'
}
```

Innders (approved):

PR #745, opened 2025-10-02
Verbose password complexity error message

### LINKED ISSUE #744: Password: password endpoints return only generic error instead of specific validation messages
## Story

When changing a password that does not meet the security requirements, the backend does not return a clear error message.  
Because of this, the frontend cannot inform the user why their password was rejected.

Currently, the backend only sends a generic message:  
`Password does not meet complexity requirements`

### To reproduce

1. Try changing the password to something invalid (e.g., too short, missing letters, etc.).
2. The request fails, but the backend responds with only the generic error message.

### Expected behavior

The backend should return clear and specific error messages (e.g.,  
`Password must be at least 8 characters long` or  
`Password must contain letters`) so the frontend can display them.

### Additional context

The issue occurs on:  
- `/api/users/passwordReset`  
- `/api/users/:user/password`  

The frontend only displays messages provided by the backend, so meaningful error messages are required for proper user feedback.

martastain:
This pull request refactors the password complexity validation logic across the codebase to simplify and unify the approach. The main change is to use the `validate_password` function directly, replacing the previous pattern of calling `ensure_password_complexity` and handling its return value or exceptions. This results in more straightforward and consistent password validation.

```
>>> from ayon_server.auth.utils import validate_password


>>> validate_password("asdf")
ayon_server.exceptions.LowPasswordComplexityException: Password must be at least 8 characters and contain digits and special characters.

>>> validate_password("asdf123")
ayon_server.exceptions.LowPasswordComplexityException: Password must be at least 8 characters and contain special characters.

>>> validate_password("asdf13@")
ayon_server.exceptions.LowPasswordComplexityException: Password must be at least 8 characters.

>>> validate_password("asdf123@")
None
```

### DIFF
--- a/ayon_server/auth/password.py
+++ b/ayon_server/auth/password.py
@@ -6,8 +6,8 @@ from ayon_server.api.clientinfo import get_real_ip
 from ayon_server.auth.session import Session, SessionModel
 from ayon_server.auth.utils import (
     create_password,
-    ensure_password_complexity,
     hash_password,
+    validate_password,
 )
 from ayon_server.config import ayonconfig
 from ayon_server.entities import UserEntity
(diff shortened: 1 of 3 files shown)


filipvnencak (approved):
Tested on Reset and Account pages,responses show the correct messages. 😉

PR #753, opened 2025-10-07
Fix: Update thumbnail ID with restricted attributes

martastain:
Fixes the following:

- Users that are allowed to publish, but their attribute write access is restricted can now update thumbnail id
- entity access check is now performed in the same transaction as operations, so the access control list is up-to-date when multiple hierarchical items are created in the same operations batch


LiborBatek (approved):
Works like charm!

LGTM

Screenshot 2025-10-07 141600


Screenshot 2025-10-07 142705

iLLiCiTiT (approved):

PR #789, opened 2025-10-31
Automatic server addon settings migration

martastain:
This pull request introduces automatic migration of server addon settings during bundle updates, refactors migration logic for better handling of annotated types, and adds a dedicated function to migrate server addon settings. The changes improve reliability and maintainability of addon settings migration, especially for production bundles.

## Summary

When a server addon version changes within a bundle, settings are automatically migrated from the original version to the new version. That includes both studio level and project level settings. 

It is however possible that reverting addon to previous version causes a loss of settings that were not available in an older version. That is expected. But it handles correctly situations such as "Ynput cloud login" and "Remote support" toggles in Ynput cloud addons, Activity categories and formulas in powerpack, entity templates etc.

## Testing notes

The new feature only affects server addons, but in order to make it work, a bunch of changes happened in the migration tools. Please confirm that copy settings (both in the editor and by bundle) works correctly, especially copying of list fields and submodels.

### DIFF
--- a/api/bundles/bundles.py
+++ b/api/bundles/bundles.py
@@ -19,7 +19,7 @@ from ayon_server.utils import RequestCoalescer
 
 from .actions import promote_bundle
 from .check_bundle import CheckBundleResponseModel, check_bundle
-from .migration import migrate_settings
+from .migration import migrate_server_addon_settings, migrate_settings
 from .models import AddonDevelopmentItem, BundleModel, BundlePatchModel, ListBundleModel
 from .router import router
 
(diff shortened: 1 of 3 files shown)


mkolar:
> I can see why the setting was marked as override although, it was empty. that's why copy settings included it even it didn't have any overrides.

Being empty is different than the default, hence it's overriden

MilaKudr (approved):
The settings automatically migrated to the new server addon version. This happened after updating the Powerpack addon in my production bundle from 1.3.0 to 1.4.0.
Screenshot 2025-11-03 163258

PR #803, opened 2025-11-11
Base views

martastain:


filipvnencak (changes requested):
The logic for loading __base__ works only when no default or working view exists.
So for artists who already have their own default view, it’s not possible to load the __base__ view through the /default endpoint — is that correct?

From what I see in views.py, the priority is:
	1.	User’s explicitly set default
	2.	User’s working view
	3.	__base__ as a fallback

This means:
	•	If a user has a working view → it loads instead of __base__
	•	If a user has a default view → it loads instead of __base__
	•	Only when neither exists → __base__ is returned
	
So in the user (artist) should get __base__ after "reseting" the view like is written here:
[#1472](https://github.com/ynput/ayon-frontend/issues/1472) after requesting GET `api/views/overview/default?project_name={project_name}`

filipvnencak:
Of the two approaches, the [GET] /base endpoint seems nicer and more intuitive to me.

filipvnencak (approved):
Working as expected, thanks!!

PR #811, opened 2025-11-18
Folder type colors

martastain:
This pull request adds support for a `color` attribute to folder types across the codebase. The changes ensure that folder types can now have an associated color, which is stored, exposed via the API, and configured in settings.


martastain:
@Innders should we provide colors for the folder types in the default anatomy preset?

PR #819, opened 2025-12-03
Fixed fetching custom list item attributes

martastain:
This pull request enhances the way entity attributes are handled and exposed in the GraphQL API, focusing on ensuring that only configured attributes are included in responses and that their types are correctly determined. The main changes involve passing attribute configuration through the GraphQL context, filtering attributes based on this configuration, and improving type resolution in attribute processing.

**GraphQL attribute configuration and filtering:**

* The `all_attrib` and `own_attrib` fields in `EntityListItemEdge` now receive the GraphQL `info` context and use it to filter and serialize attributes according to the configured list of attributes. Only attributes present in the configuration are included in the output.
* The resolver `get_entity_list_items` now extracts the attribute definitions from the root entity's data and stores them in the GraphQL context as `list_attributes`, mapping attribute names to their types.

### DIFF
--- a/ayon_server/graphql/nodes/entity_list.py
+++ b/ayon_server/graphql/nodes/entity_list.py
@@ -45,7 +45,7 @@ class EntityListItemEdge(BaseEdge):
     _user: strawberry.Private[UserEntity]
 
     @strawberry.field()
-    def all_attrib(self) -> str:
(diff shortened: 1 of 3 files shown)


Innders (approved):
Seems to work

PR #822, opened 2025-12-05
Improved folder filtering

### LINKED ISSUE #821: GraphQL: Filter folders by tasks
add taskFilter to folders graphql resolver

martastain:
This pull request adds support for filtering folders by associated tasks in the `get_folders` GraphQL resolver. The main change introduces a new `task_filter` argument, allowing clients to specify task-based criteria for folder retrieval. The implementation parses the filter, constructs a SQL Common Table Expression (CTE) to select relevant folders, and joins this with the folders query, ensuring only folders with matching tasks are returned.

image

## Folder search endpoint

Added new endpoint `[POST] /api/projects/{projectname}/folders/search` that enables searching for folder ids matching given criteria (taskFilter, folderFilter, taskSearch, folderSearch). mimicking graphql folder resolver. This endpoint deprecates taskFolders endpoint that does the same, but with limited functionality

image

### DIFF
--- a/api/folders/__init__.py
+++ b/api/folders/__init__.py
@@ -1,4 +1,4 @@
-__all__ = ["router", "folders", "list_folders"]
+__all__ = ["router", "folders", "list_folders", "search"]
 
-from . import folders, list_folders
+from . import folders, list_folders, search
 from .router import router
--- a/api/folders/search.py
+++ b/api/folders/search.py
@@ -1,0 +1,215 @@
+from typing import Annotated
+
+from ayon_server.access.utils import folder_access_list
+from ayon_server.api.dependencies import CurrentUser, ProjectName
+from ayon_server.exceptions import BadRequestException
+from ayon_server.lib.postgres import Postgres
+from ayon_server.sqlfilter import QueryFilter, build_filter
+from ayon_server.types import Field, OPModel
+from ayon_server.utils import SQLTool, slugify
+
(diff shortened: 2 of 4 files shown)


Innders:
I just realised that we don't use folders graphql resolver for the products page but we use `/folders` endpoint. Can this also be added an argument to that endpoint as well?

martastain:
> I just realised that we don't use folders graphql resolver for the products page but we use /folders endpoint. Can this also be added an argument to that endpoint as well?

Unfortunatelly no. /folders endpoint uses cached hierarchy so only filtering that happens there is for ACL. The endpoint does not use SQL at all, so we cannot use SQLFilter.

Innders (changes requested):
When the conditions are empty for tasks filter and there is a folder filter, there is an error. It doesn't happen when the other way round or when taskFilter is `undefined`.

```json
{
    "taskFilter": {
        "conditions": []
    },
    "taskSearch": "",
    "folderFilter": {
        "conditions": [
            {
                "key": "attrib.priority",
                "value": [
                    "urgent"
                ],
                "operator": "in"
            }
        ]
    },
    "folderSearch": ""
}
```

```json
{
    "code": 400,
    "detail": "Task filter provided but could not be parsed"
}
```

martastain:
Sorry. Fixed

Innders (approved):

PR #867, opened 2026-02-26
Expose `data` for the LinkEdge GraphQL response

BigRoy:
## Description of changes

Expose the `data` field to GraphQL for LinkEdge.

So that e.g.

```graphql
query GetProjectLinksData {
  project(name: "projectname") {
    folders {
      edges {
        node {
          links {
            edges {
              data
              name
            }
          }
        }
      }
    }
  }
}
```

Also provides back the `data` that was stored with the links.

### DIFF
--- a/ayon_server/graphql/nodes/common.py
+++ b/ayon_server/graphql/nodes/common.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import Any, Optional
 
 import strawberry
+from strawberry.scalars import JSON
 
 from ayon_server.entities import UserEntity
 from ayon_server.exceptions import AyonException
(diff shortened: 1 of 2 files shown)


martastain (approved):
Works like a charm. using JSON scalar is a little inconsistent with the rest of the API (that was created before this was widely available), but it is a great P.O.C. for the planned GraphQL fields redesign. I hope we'll see it soon acrorss the board

BigRoy:
> Works like a charm. using JSON scalar is a little inconsistent with the rest of the API (that was created before this was widely available), but it is a great P.O.C. for the planned GraphQL fields redesign. I hope we'll see it soon acrorss the board

I looked into the `json_dumps` approach first and returning `str | None` like the other areas first. But this worked, and I figured that it worked really nicely. ❤️ Whether it'll bite us later, we'll see 🗡️

PR #874, opened 2026-03-13
Track original bundle state in bundle.updated event

### LINKED ISSUE #839: Events: Upon setting a new active bundle variant add the old active variant to the event payload
## Story

To be able to easier trace back what a previous bundle was before a recent active bundle change, it'd be nice if the payload of the bundle change event would contain the "oldValue" of that particular bundle variant.

### Problems

Easier tracing back to what a previous bundle was at a certain point.

### Proposal

Add it to the payload data?

## TODO

- [ ] ...

martastain:
This pull request refactors the event dispatch logic in the `update_bundle` function to provide more detailed information about bundle state changes. It introduces tracking of previous bundle states and ensures that these are included in the event summary, improving traceability for updates.

Enhancements to event dispatch and bundle state tracking:

* Added previous state of the bundle (`wasProduction`, `wasStaging`, `wasArchived`) to the event summary

image

### DIFF
--- a/api/bundles/bundles.py
+++ b/api/bundles/bundles.py
@@ -288,6 +288,10 @@ async def update_bundle(
             is_archived=row["is_archived"],
         )
 
+        was_production = bundle.is_production
+        was_staging = bundle.is_staging
+        was_archived = bundle.is_archived
+
         if patch.is_archived and (bundle.is_production or bundle.is_staging):
             raise BadRequestException(
                 "Cannot archive bundle that is production or staging"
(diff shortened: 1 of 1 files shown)


BigRoy (commented):
I personally preferred the `newValue` and `originalValue` payload that settings changes e.g. have because it's less duplication of `is` and `was` entries. 

Also, I'm not sure this event really gives me the input I'd really be looking for quickly. Because what I'd really love to actually know is... that if this bundle is set to production - what was the previous production bundle: `None` or `old-bundle-name`.

🤔 

However, other than that - this does work fine, however it does provide relatively limited data that I think you're often not too concerned about.

PR #876, opened 2026-03-16
Add isEmpty flag to project folders

### LINKED ISSUE #875: YN-0581: Project folders: Hide empty folders with restricted projects

Empty folders are not useful for other users, especially when they are empty because the projects inside are hidden due to permissions.

When a folder is **not** flagged is isEmpty but it is empty for the user, this means it has restricted projects inside of it. These folders should be hidden. 

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1480830290858606613/1480854692861841451)

_This issue was automatically created from AYON ticket [YN-0581]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=9aedb3b0212211f194f38df40cfacb4a)

martastain:
This pull request introduces an enhancement to the project folders API by adding the ability to determine whether a folder (or any of its subfolders) contains projects. The main change is the addition of an `is_empty` field to the `ProjectFolderModel`, along with logic to compute this value for each folder.

Enhancement to project folder metadata:

* Added an `is_empty` boolean field to the `ProjectFolderModel` to indicate whether the folder (or any of its subfolders) contains any projects.

Logic for computing folder emptiness:

* Implemented a Python-based recursive crawl in `get_project_folders` to mark folders as empty if they and all their subfolders contain no projects, using a set of folder IDs derived from the projects table.

### DIFF
--- a/api/project_folders/project_folders.py
+++ b/api/project_folders/project_folders.py
@@ -60,6 +60,12 @@ class ProjectFolderModel(OPModel):
     label: Annotated[str, FFolderLabel]
     parent_id: Annotated[str | None, FFolderParentID] = None
     position: Annotated[int, Field(title="Folder position", ge=0)] = 0
+    is_empty: Annotated[
+        bool,
+        Field(
+            title="Whether the folder (or any of its subfolders) contains any projects",
+        ),
+    ] = False
     data: Annotated[ProjectFolderData, FFolderData]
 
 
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Works as expected.

Innders:
Can we merge then?

PR #877, opened 2026-03-16
Feature: Data export and imports

### LINKED ISSUE #833: YN-0350: csv bulk import / update

full csv import /update feature to bulk create hierarchies, assignments, etc. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0350]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=34faf7c0ebc611f0b08ac15dfa6bf60d)

kalisp:
This PR contains functionality for export and imports of various entities into CSV files.

Currently implemented:
- users
- folders
- tasks
- folder hierarchy

It supports selection of exportable fields, behavior of imports (fail, skip errors) etc.

### Technical details

Moved out of ayon-power-pack https://github.com/ynput/ayon-power-pack/pull/58 as it should be part of server backend.


api/data_import/import_data.py
```diff
@@ -0,0 +1,338 @@
...
+            if skip_errors:
+                continue
+            raise exp
+
+        imported += 1
+
+    return imported
+
+
+def _get_entity_class(entity_type: str) -> type:
```
martastain: There is `get_entity_class` function in `ayon_server.helpers.get_entity_class` function that does exactly this :)

api/data_import/router.py
```diff
@@ -0,0 +1,3 @@
+from fastapi import APIRouter
+
+router = APIRouter(tags=["Data Import"], prefix="/data_import")
```
martastain: `/data_import` entrypoint will produce routes such as `/api/data_import/export` - wouldn't it make sense to use prefix `/csv`, so the final endpoints will be `/api/csv/import` , `/api/csv/export` etc? Also - we'd avoid snake/camel/kebab case confusions

api/data_import/import_data.py
```diff
@@ -0,0 +1,338 @@
...
+}
+
+HIERARCHY_ENTITY_CLASSES: dict = {
+    "folder": FolderEntity,
+    "task": TaskEntity,
+}
+
+
+@router.post("/import/{entity_type}")
+async def import_data(
```
martastain, line 143: Before actual ACL is decided, please add:

`user: CurrentUser` argument to all endpoints (from `ayon_server.api.dependencies`) and do `if not user.is_manager: raise ForbiddenException("you must be a manager")` check to prevent unauthenticated users and artists from dumping and updating data.

api/data_import/import_data.py
```diff
@@ -0,0 +1,338 @@
...
+    entity_type: str,
+    file_bytes: bytes,
```
martastain: Do you expect more complex arguments here? File uploads should probably handled either as multiparts, or as "file in body, argumens in query arguments" method - then you can use `request.body()` to get the actual bytes. In both cases, `Annotated` should be used for arguments so the generated openapi.json contains all the constraints:

```
ImportEntityType = Literal["folder", "user", "task", "hierarchy"]

entity_type: Annotated[ImportEntityType, fastapi.Path(title="Import entity type")]
```

api/data_import/models.py
```diff
@@ -0,0 +1,518 @@
+"""Data models for data import/export functionality."""
+
+from __future__ import annotations
```
martastain: We already live in the future :)

api/data_import/models.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+class ExistingItemStrategy:
+    """Strategy for handling existing items during import."""
+    SKIP = "skip"
+    UPDATE = "update"
+    FAIL = "fail"
+
+
+class EntityExportImport(BaseModel):
```
martastain: I am maybe missing something, but why is this a pydantic model?

api/data_import/models.py
```diff
@@ -0,0 +1,518 @@
...
+from typing import Any, Dict, List, Tuple, Optional, Set
+from pydantic import BaseModel, Field
+from pydantic.fields import ModelField, FieldInfo
+
+from ayon_server.entities import UserEntity, FolderEntity, TaskEntity
+from ayon_server.lib.postgres import Postgres
+from ayon_server.types import OPModel
+
+
+class ExistingItemStrategy:
```
martastain: Shouldn't this be an enum or Literal, so it could be used for validation?

api/data_import/export_data.py
```diff
@@ -0,0 +1,84 @@
...
+        entity_ids=entity_ids
+    )
+
+    export_path = os.path.join(STORAGE_DIR, f"{entity_type}_export.csv")
+    with open(export_path, "w", encoding="utf-8", newline="") as f:
+        writer = csv.writer(f)
+        writer.writerows(rows)
+
+    return FileResponse(
+        export_path,
```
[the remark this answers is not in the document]
martastain: if entity_type was a Literal, copilot wouldn't complain :)

api/data_import/import_data.py
```diff
@@ -0,0 +1,338 @@
...
+    if semicolon_count > comma_count:
+        return ";"
+    return ","
+
+
+async def _get_entity_id_by_path(
+    project_name: str,
+    path: str,
+    is_task: bool = False
+) -> Optional[str]:
```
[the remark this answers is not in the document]
martastain: I'd say it should raise NotFoundException or similar.

api/data_import/models.py
kalisp: Removed

Innders (approved):
Works for me, can we get this merged @martastain ?

martastain (changes requested):
@kalisp  Could you please use `make check` to run ruff and mypy and fix the style and type issues? 
There's also a pre-commit hook that can be installed using `make`.

api/data_import/common.py
```diff
@@ -0,0 +1,61 @@
+from ayon_server.exceptions import NotFoundException
+from ayon_server.lib.postgres import Postgres
+
+# Sender type for operations status propagation
+SENDER_TYPE = "data_import"
+
+
+async def _get_entity_id_by_path(
```
martastain: Underscore denotes a private function, but it is not used in the file and is importd from import_data, so technically it shouldn't start with an underscore.

api/data_import/import_data.py
```diff
@@ -0,0 +1,1018 @@
...
+        try:
+            if import_type == "entity_list_item":
+                entity_cls: type[Any] = EntityListItemModel
+            elif import_type == "hierarchy":
+                entity_type = await _get_entity_type(
+                    project_name, row, column_mapping, fields
+                )
+                if entity_type not in HIERARCHY_MODEL_CLASSES:
+                    error_msg = f"Invalid entity_type '{entity_type}'"
+                    raise ValueError(error_msg)
```
martastain: Please use BadRequestException for this kind of errors. ValueError (and any exception that is not a child of AyonException is treated as internal server error, but in this case it is a user error. 

Internal server errors are logged in the event log with full traceback, whereas BadRequestException "just" return the error message to the user, so they can fix it without flooding logs

api/data_import/export_data.py
```diff
@@ -0,0 +1,94 @@
...
+from .router import router
+
+# Type alias for exportable entity types
+EntityType = Literal["user", "folder", "task", "hierarchy", "entity_list_item"]
+
+
+@router.get("/export/{entity_type}/fields")
+async def export_fields(
+    entity_type: Annotated[EntityType, fastapi.Path(title="Import entity type")],
+    project_name: str | None = None,
```
martastain: It is not very obvious where the project_name is injected from. If it is a query argument, pls use

```python
project_name: Annotated[str | None, fastapi.Query(alias="project_name", regex=PROJECT_NAME_REGEX) = None

and then:

```python
from ayon_server.helpers.project_list import normalize_project_name

if project_name is not None:
    project_name = await normalize_project_name(project_name)
```

Which ensures the project actually exists and also normalizes the leter casing in the case it is provided incorrectly

ayon_server/settings/anatomy/__init__.py
```diff
@@ -24,7 +24,7 @@
 from ayon_server.settings.anatomy.templates import Templates
 from ayon_server.settings.common import BaseSettingsModel
 from ayon_server.settings.settings_field import SettingsField
-from ayon_server.settings.validators import ensure_unique_names
```
martastain, line 27: I am not sure this is a good idea: I believe ensure_unique_names is a validator already used in addons. ensure_unique_property is fine, but we need to keep the original as well (that would under the hood just call ensure_unique_property)

ayon_server/helpers/get_entity_class.py
```diff
@@ -46,6 +47,7 @@ def get_entity_class(entity_type: str) -> type[ProjectLevelEntity]:
         "version": VersionEntity,
         "representation": RepresentationEntity,
         "workfile": WorkfileEntity,
+        "user": UserEntity,
```
martastain: I wonder why mypy does not complain about this? get_entity_class is supposed to return `type[ProjectLevelEntity]` which UserEntity is not. 

I guess the correct way would be to annotate the return type as 
`type[ProjectLevelEntity] | type[UserEntity] and then add correct overloads for the function, but...screw that for now. It will complain as soon we turn the strict mode on and then we can fix it. Maybe just add a little note there?

api/data_import/export_data.py
```diff
@@ -0,0 +1,94 @@
...
+from fastapi.responses import FileResponse, Response
+
+from ayon_server.api.dependencies import CurrentUser
+from ayon_server.exceptions import ForbiddenException
+
+from .models import EXPORTABLE_ENTITIES, ImportableColumn
+from .router import router
+
+# Type alias for exportable entity types
+EntityType = Literal["user", "folder", "task", "hierarchy", "entity_list_item"]
```
martastain: This should probably go to common or models  .... also EntityType feels very generic to me. ImportEntityType maybe to avoid confusions?

api/data_import/import_data.py
```diff
@@ -0,0 +1,1018 @@
...
+        raise BadRequestException("Invalid content type")
+    file_id = create_uuid()
+    ttl_seconds = ttl if ttl is not None else 30 * 60  # 30 minutes default
+    await Redis.set(REDIS_NS, file_id, csv, ttl=ttl_seconds)
+
+    return ImportUpload(id=file_id)
+
+
+@router.post("/import/{import_type}")
+async def import_data(
```
martastain, line 143: Not critical, but I'd prefer using Annotated with explicit `Query` field to denote that the parameter comes from the query parameter - it also allows you to add regex validators for string fields easily.

api/data_import/common.py
martastain: Also. This is called for every row, which is very resource heavy. Does it need to take into account entities that have been "just created" by a previous row? couldn't the hierarchy be loaded once and then updated during the iteration if needed? That would require a major refactoring, so let's keep it this way for now, but it would be good to do that as a future update - we have folder cache in redis, so it could be loaded once when the import starts?

martastain (changes requested):

api/data_import/common.py
kalisp: It does uses cache before calling this method.

Renamed.

api/data_import/export_data.py
kalisp: Extracted to common

ayon_server/settings/anatomy/__init__.py
kalisp: Returned back.

ayon_server/helpers/get_entity_class.py
kalisp: Added todo

api/data_import/export_data.py
kalisp: Extracted to common.

api/data_import/import_data.py
kalisp: Didnt understand. What should be Annotated?

api/data_import/import_data.py
kalisp: Replaced all ValueError with BadRequestException

martastain (approved):

PR #893, opened 2026-03-27
Feature: Added TagsEnumResolver

kalisp:
## Description of changes

Added `TagsEnumResolver` to `enum_anatomy` to other existing enums as `StatusesEnumResolver`, `FolderTypesEnumResolver`

### DIFF
--- a/ayon_server/enum/resolvers/enum_anatomy.py
+++ b/ayon_server/enum/resolvers/enum_anatomy.py
@@ -131,7 +131,6 @@ class TagsEnumResolver(BaseEnumResolver):
         return {"project_name": str}
 
     async def resolve(self, context: dict[str, Any]) -> list[EnumItem]:
-        print(f"resolve::{context}")
         project_name = context.get("project_name")
         if not project_name:
             anatomy = await get_primary_anatomy_preset()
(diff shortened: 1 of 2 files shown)


ayon_server/enum/resolvers/enum_anatomy.py
```diff
@@ -122,3 +122,41 @@ async def resolve(self, context: dict[str, Any]) -> list[EnumItem]:
...
+
+
+class TagsEnumResolver(BaseEnumResolver):
+    name = "tags"
+
+    async def get_accepted_params(self) -> dict[str, type]:
+        return {"project_name": str}
+
+    async def resolve(self, context: dict[str, Any]) -> list[EnumItem]:
+        print(f"resolve::{context}")
```
martastain: rogue print :)

martastain:
It is good to have it, but I'm thinking if this is strictly needed for the import - tags are not enforced and you can add a tag to an entity even if it is not a part of the project anatomy - it is just not colored.

ayon_server/enum/resolvers/enum_anatomy.py
kalisp: removed

kalisp:
Merged with https://github.com/ynput/ayon-backend/pull/877

PR #896, opened 2026-03-30
Implemented adding new enum items

kalisp:
## Description of changes

This PR adds new method to `BaseEnumResolver` and implements it for enums where it makes sense.

This provides at least some api for creation of new enum items, as the logic/format/destination might be different for each of the enums.

### Additional context

Logic in `enum_anatomy` could be refactored, maybe added as default behavior to `BaseEnumResolver` but that didn't seem right as current implemention is field tested by `data_import`.

### DIFF
--- a/ayon_server/enum/base_resolver.py
+++ b/ayon_server/enum/base_resolver.py
@@ -1,7 +1,6 @@
 from typing import TYPE_CHECKING, Any
 
 from ayon_server.forms import SimpleForm
-from ayon_server.types import SimpleValue
 
 from .enum_item import EnumItem
 
@@ -36,7 +35,7 @@ class BaseEnumResolver:
         item: EnumItem,
         project_name: str | None = None,
         **kwargs,
-    ) -> SimpleValue:
+    ) -> None:
         _ = (
             item,
             project_name,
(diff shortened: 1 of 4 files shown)


martastain (approved):
The code looks good, i took the liberty of fixing a few type hints to make mypy happy.

I have two concerns, not sure if blocking:

What's the point of returning the value of the item? It is already passed in vie EnumItem itself, right? I thought it is because auto-generated values, such as database IDs, but item.value must be present in EnumItem so you are always creating an explicit value.

In the link types resolver, kwargs contain elements that are very specific to the created link type, but not at all to the EnumItem itself. Shouldn't we wrap it to a custom kwarg like this?

```python

create_item("linkTypes", item=my_link_type, project_name="my_project", link_type_params={"input_type": "folder", "output_type": "folder})
```
And keep the top-level kwargs for context specifications? Just an idea, not forcing it.

I wasn't able to test the functionality, but i assume, you are able to test it with csv import code better than me

kalisp:
> What's the point of returning the value of the item

Yeah, initial idea was to return auto-generated id, but it seems that we are not using it for enum item. I'll remove return value as it is unnecessary.

kalisp:
> In the link types resolver

Makes sense, I didn't fancy that implementation either, this looks better. Updated.

PR #897, opened 2026-03-30
Feature: Team mentions

kalisp:
Implemented endpoint to enrich list of offered users with team names when `@NAME` is used in comment.

image

### Additional context

This mention is parsed when comment is created and separate references are created for each team member.

It doesn't change mention format from :user to :team as it needs update on frontend side.

This is fairly naive implementation without requested future proofing as creating new model etc.

WIP


martastain:
The logic seems to work nicely, team is correctly expanded to users. As you mentioned, you are now pushing it to "users" field of the suggest response because teams is not supported in the frontend, so backend and frontend should be merged at the same time and team markdown tag shouldn't hijack "user" entity type.

martastain (changes requested):
Merging is blocked by the frontend and switching to updated suggest model.

filipvnencak:
@kalisp Can you add new team  array to `/api/projects/{Project_name}/suggest` so teams will be in the own array not in the users array?

Innders (changes requested):
I can't seem to get the notifications to work.

https://github.com/user-attachments/assets/017af329-3f52-4645-a30a-6b9e00d34011

martastain:
> I can't seem to get the notifications to work.

confirmed. extract mentions still uses "user" tag, but the mention is now correctly set to team, @kalisp 

```python

async def extract_link_tuples(md_text: str, project_name: str) -> list[EntityLinkTuple]:
    links: set[EntityLinkTuple] = set()
    team_names = set(
        team.name for team in
        await get_team_suggestion_items(project_name)
    )
    for link in LINK_PATTERN.findall(md_text):
        try:
            entity_type, entity_id = link[1].split(":")
            if entity_type not in get_args(ReferencedEntityType):
                continue
            if entity_type == "user" and entity_id in team_names:
                members = await get_team_members(project_name, entity_id)
                for member_name in members:
                    links.add(("user", member_name))
            else:
                links.add((entity_type, entity_id))
        except ValueError:
            logger.debug("Invalid reference link format")
    return list(links)

```

kalisp:
Updated parsing of mention string.

martastain (changes requested):
It still doesn't seem to expand team mention to list of users :-(

kalisp:
Because I failed to push it, skill issue :/.

Innders (approved):
Working now thanks 💯

kalisp:
What would be a next steps here in regards of attached PR for frontend? Which should be merged first, where should be set dependency?

Innders:
@martastain can you please approve and merge?

martastain (approved):

PR #906, opened 2026-04-13
Extended text search in graphql

### LINKED ISSUE #909: YN-0648: Search for multiple folders at the same time

Can we copy and paste shots into the search bar and find only those shots. 

We'd like to be able to paste a list of strings into the search field and filter out all of those shots or tasks based on the name.

<br>

We should allow to break a text with deliminer (comma or space?) into individual OR search strings.

### Additional context:

_This issue was automatically created from AYON ticket [YN-0648]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=c3e4b3d0329411f1916bcf676d7f83ef)

martastain:
This pull request improves the search functionality in both the `get_folders` and `get_tasks` GraphQL resolver functions. The main change is to support searching with multiple comma-separated terms, making the search more flexible and powerful.

Search string is split by a coma and those parts are evaluated using OR operator, and each part is slugified and evaluated using AND operator. That extends the search, so it is possible to search for multiple folders/tasks at once while maintaining 100% backwards compatibility

image

### DIFF
--- a/ayon_server/graphql/resolvers/tasks.py
+++ b/ayon_server/graphql/resolvers/tasks.py
@@ -431,16 +431,21 @@ async def get_tasks(
 
     if search:
         use_folder_query = True
-        terms = slugify(search, make_set=True)
-        # isn't it nice that slugify effectively prevents sql injections?
-        for term in terms:
-            cond = f"""(
-            tasks.name ILIKE '%{term}%'
-            OR tasks.label ILIKE '%{term}%'
-            OR tasks.task_type ILIKE '%{term}%'
-            OR hierarchy.path ILIKE '%{term}%'
-            )"""
-            sql_conditions.append(cond)
+        parts = search.split(",")
+        t1_conds = []
+
+        for part in parts:
+            terms = slugify(part, make_set=True)
+            t2_conds = []
+            for term in terms:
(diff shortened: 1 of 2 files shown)


ayon_server/graphql/resolvers/tasks.py
```diff
@@ -431,16 +431,21 @@ async def get_tasks(
...
-            OR hierarchy.path ILIKE '%{term}%'
-            )"""
-            sql_conditions.append(cond)
+        parts = search.split(",")
+        t1_conds = []
+
+        for part in parts:
+            terms = slugify(part, make_set=True)
+            t2_conds = []
+            for term in terms:
```
filipvnencak, line 440: in folders is `term = term.replace("'", "''")` here is missing, there is slugify() for safety, but isn't it be consistent??

filipvnencak (commented):
Working perfectly on fronted

ayon_server/graphql/resolvers/tasks.py
martastain: ah. good catch. yeah. that's totally redundant. removed

filipvnencak (approved):

PR #915, opened 2026-04-16
Project skeletons, labels and thumbnails

### LINKED ISSUE #879: Project labels
Set a nice looking label for the project that is purely cosmetic and can be changed at anytime.

This issue is linked to our feedback platform. For more details and updates, please visit [this link](https://ayon.featurebase.app/p/project-nice-namelabel).

### LINKED ISSUE #880: Project thumbnails
Upload thumbnails to projects just like other entities. These thumbnails can be used in the project list to better identify different projects.

This issue is linked to our feedback platform. For more details and updates, please visit [this link](https://ayon.featurebase.app/p/project-thumbnails).

martastain:
## Project skeleton

This pull request introduces significant improvements to project and team management APIs, with a focus on supporting "project skeletons" (minimal project structures), enhanced project creation options, and more consistent API dependency handling. Key changes include allowing operations on skeleton projects, expanding the project deployment API, and improving project listing and team management endpoints.

* Added `AllowProjectSkeleton` dependency to relevant endpoints, allowing operations (such as GET, POST, PATCH, DELETE) on skeleton projects for resources like projects, users, teams, and thumbnails. This ensures that skeleton projects are handled consistently across the API.
* Expanded `deploy_project` API to support creating either a full project or a skeleton project, with the option to use an anatomy preset. Added validation to prevent duplicate project names/codes and improved error handling for missing anatomy or presets. 
* Enhanced project listing to support filtering by skeleton status, and included new fields (`label`, `skeleton`) in the response model. Improved SQL query and result parsing for these fields. 
* Refactored team and team member endpoints to use path parameters instead of query parameters for `team_name` and `member_name`, and added `AllowProjectSkeleton` dependency to all relevant endpoints for consistency.

## Project thumbnails

Added support for project entity thumbnails via `[POST/GET] /api/projects/{project_name}/thumbnail`

Standard project thumbnails are stored the same way project level entities (original on the project storage, scaled down version in the thumbnails table in the project schema), skeleton projects only store scaled down versions in public.project_thumbnails table.

## Renaming projects

It is now possible to rename a project using `[PATCH] /api/projects/{projectName}/rename` with `newName` field set.

Limitations:

- Warning: This is highly experimental and might not cover everything
- It is not possible to rename a project with media files (thumbnails, reviewables, comment uploads)

## Project labels

Project now can have optional labels that could be displayed instead of the machine-friendly names. Labels are optional and when unset, displayed title should fallback to the project name

image


ayon_server/entities/project.py
```diff
@@ -361,3 +406,8 @@ def link_types(self) -> Sequence[LinkTypeModel]:
     def link_types(self, value: list[dict[str, Any]]) -> None:
         """Set the link types."""
         self._payload.link_types = value  # type: ignore
+
+    @property
+    def skeleton(self) -> bool:
+        """Return True if the project is a skeleton."""
+        return False
```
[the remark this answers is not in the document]
martastain: that's correct.

Innders (changes requested):
- [x] `color` field on project.
- [ ] Expose project data fields to list projects REST endpoint.
- [x] When not providing an anatomy preset name, use the default (primary) preset.
- [x] Deploy Project endpoint should let you set `label` and `color`
- [x] Deleting skeleton projects does not work.
```json
{
    "status": 500,
    "detail": "TypeError: Type is not JSON serializable: asyncpg.protocol.record.Record",
    "traceback": "TypeError: Type is not JSON serializable: asyncpg.protocol.record.Record\n\nayon_server/api/context.py:77\nresponse = await call_next(request)\n\napi/projects/projects.py:251\nproject = await ProjectEntity.load(project_name)\n\nayon_server/entities/project.py:103\nawait Redis.set_json(\n\nayon_server/lib/redis.py:153\npayload = json_dumps(value)\n\nayon_server/utils/json.py:51\nreturn orjson.dumps("
}
```

- [ ] GetProject by name does not work for skeleton projects.

martastain:
color is implemented as `data.color`, exposed in graphql as project top level field
project list will not contain all project data. it would be huge.

ayon_server/entities/project.py
```diff
@@ -361,3 +408,8 @@ def link_types(self) -> Sequence[LinkTypeModel]:
...
```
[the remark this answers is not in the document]
martastain: this is the correct behavior. (i think)

maintenance/tasks/remove_unused_thumbnails.py
```diff
@@ -32,6 +32,8 @@ async def clear_thumbnails(project_name: str) -> None:
 
     storage = await Storages.project(project_name)
     async for row in Postgres.iterate(query):
+        if row["id"] == "0" * 32:
+            continue
```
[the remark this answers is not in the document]
martastain: nope. it doesn't it uses EntityID.parase both for encoding and decoding

ayon_server/helpers/rename_project.py
```diff
@@ -0,0 +1,102 @@
...
+        raise BadRequestException(
+            f"Project {project_name} has thumbnails, cannot rename."
+        )
+
+    res = await Postgres.fetchrow(f"SELECT COUNT(*) FROM project_{project_name}.files")
+    if res and res[0] > 0:
+        raise BadRequestException(f"Project {project_name} has files, cannot rename.")
+
+
+async def rename_project(
```
iLLiCiTiT, line 46: We probably should validate if the new project already exists?

ayon_server/helpers/thumbnails.py
```diff
@@ -222,3 +224,69 @@ async def store_thumbnail(
...
+    meta = {
+        "originalSize": len(payload),
+        "thumbnailSize": len(thumbnail),
+        "mime": mime,  # eventually, we'll drop the column
+    }
+    if user_name:
+        meta["author"] = user_name
+
+    query = """
+        INSERT INTO public.project_thumbnails (project_name, mime, data, meta)
```
iLLiCiTiT: If we store only skeleton project names there it should be `public.project_skeleton_thumbnails`.

ayon_server/helpers/rename_project.py
martastain: The transaction would fail with a conflict exception (if either name or code already exists) and would be rolled back

api/projects/projects.py
```diff
@@ -250,10 +263,60 @@ async def delete_project(
...
+    return None
+
+
+#
+# Rename
+#
+
+
+class RenameProjectRequestModel(OPModel):
+    new_name: Annotated[
```
iLLiCiTiT: Annoying question: Does it make sense to have `new_` prefix in the data? The request is "rename" project, so I would just expect a `name` and `code` instead of `new_name` and `new_code`.

martastain: you're right. i wanted to keep it consistent with rename user endpoint, where no one stopped me from naming it stupidly, but i shouldn't follow old patterns when they are bad. will fix

ayon_server/helpers/rename_project.py
iLLiCiTiT: Yep, tried and it fails.

iLLiCiTiT:
How can I create skeleton project?

martastain:
> How can I create skeleton project?

Just add `skeleton:true` to `[POST] /api/projects` request

iLLiCiTiT (approved):
Looks like all works as expected. The `new_` prefix is something I would change.

Skeleton project do exist when are created, and are not listed via graphql if not specifically asked for.

iLLiCiTiT:
One more annoying question: The filter for graphql is `skeleton` which is `bool`, when set to `true` it returns both standard and skeleton projects. Maybe it should be `includeSkeleton` instead?

martastain:
> One more annoying question: The filter for graphql is skeleton which is bool, when set to true it returns both standard and skeleton projects. Maybe it should be includeSkeleton instead?


Fair point, will fix

PR #922, opened 2026-04-28
BaseTypes:  Filter and grouping

filipvnencak:
## PR Checklist

## Description of changes

Adds `productBaseType` as a top-level grouping key for entity groups. Versions GraphQL resolver now exposes `product_base_type` as a virtual/queryable field joined from the `products` table.

### Technical details

- New resolver `get_product_base_type_groups` in `ayon_server/entities/grouping/resolvers.py`. Returns one group per base type defined in project anatomy (count=0 if unused), plus orphan values found in `products.product_base_type` that aren't in the anatomy. Icon/color sourced from anatomy definitions, falling back to `product_base_types.default`.
- `api/grouping/grouping.py`: registered `productBaseType` → `product_base_type` in `TOP_LEVEL_GROUPING_KEYS` and routed to the new resolver.                                      
- `ayon_server/graphql/resolvers/versions.py`: added `product_base_type` to allowed columns, column_map (`products.product_base_type`), and product subselect — mirrors existing `product_type` handling.

### Additional context
Frontend [#1939](https://github.com/ynput/ayon-frontend/pull/1939)

### DIFF
--- a/api/grouping/grouping.py
+++ b/api/grouping/grouping.py
@@ -11,6 +11,7 @@ from ayon_server.entities.grouping.common import EntityGroup
 from ayon_server.entities.grouping.resolvers import (
     get_assignees_groups,
     get_attrib_groups,
+    get_product_base_type_groups,
     get_product_type_groups,
     get_status_or_type_groups,
     get_tags_groups,
@@ -27,6 +28,7 @@ TOP_LEVEL_GROUPING_KEYS = {
     "taskType": "task_type",
     "folderType": "folder_type",
     "productType": "product_type",
+    "productBaseType": "product_base_type",
     "assignees": "assignees",
     "status": "status",
     "tags": "tags",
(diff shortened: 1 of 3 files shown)


martastain (changes requested):
The logic for product base type is that if it is not set, product type should be used - this is to provide backwards compatibility with older projects, so at least temporarily we should take null values in the database into account and use fallback to product type if the project is not fully migrated:

maybe something like this?

```sql
        WITH base_types_with_fallback AS (
            SELECT
                COALESCE(product_base_type, product_type) AS product_base_type
            FROM project_{project_name}.products
        )
        SELECT count(*) AS count, product_base_type AS value
        FROM base_types_with_fallback
        GROUP BY product_base_type
```

Otherwise the code looks good and works.

martastain (approved):

PR #923, opened 2026-04-28
fix: match graphql search with folder_search endpoint

martastain:


api/folders/search.py
```diff
@@ -118,15 +118,21 @@ async def search_folders(
...
+                t2_conds = []
+                for term in terms:
+                    t2_conds.append(
+                        f"(tasks.name ILIKE '%{term}%'"
+                        f"OR tasks.label ILIKE '%{term}%'"
+                        f"OR tasks.task_type ILIKE '%{term}%'"
+                        f"OR ex.path ILIKE '%{term}%')"
+                    )
+                t1_conds.append(SQLTool.conditions(t2_conds, "AND", add_where=False))
+            sql_conditions.append(SQLTool.conditions(t1_conds, "OR", add_where=False))
```
filipvnencak: This returning 
 ```
{                                                                                                                                                                                    
      "status": 500,                                                                                                                                                                                
      "detail": "asyncpg.exceptions.UndefinedTableError: missing FROM-clause entry for table \"tasks\"",                                                                                            
      "traceback": "asyncpg.exceptions.UndefinedTableError: missing FROM-clause entry for table \"tasks\"\n\napi/folders/search.py:213\nasync for row in Postgres.iterate(query):\n\nayon_server/lib/postgres.py:295\nstatement = await conn.prepare(query)\n\nasyncpg/protocol/protocol.pyx:165"                                                          
  } 
```

filipvnencak (changes requested):
I'm also getting 19 times same task the graphql call 19 times and ended up that I have 19 modeling in one folder, but it is only for second and n searching chip 00_ploeptaemp lookdev, 01_mcleutsunds modeling is my search input, for lookdev I have correct one task, but for modeling on 01...folder I have duplicates

Image

PR #933, opened 2026-05-15
Activity feed filter

### LINKED ISSUE #240: Activity Feed: Filter activities by search query
## Story

I want to search for a specific message or only show messages from a specific author.

### Problems

- How can different fields be searched like **user** and **createdAt**

### Proposal

- New graphql argument `query`
- Query language like github. `is:comment author:luke mention:felix after:12-06-24`

We will need to have query filtering like this in many places, tasks, inbox, events. It might be worth getting it write the first time.

kalisp:
## Description of changes

Added new `filter` argument for detailed filtering of `activity_feed` table similar to other graphQl resolvers.

### Technical details

Provides graphQl filter like:
```
{"filter": "{\"conditions\":[{\"key\": \"activity_data.author\", \"operator\":\"eq\", \"value\":\"admin\"}]}"}
```

Existing arguments kept for backward compatibility and simpler usage.

### DIFF
--- a/ayon_server/graphql/resolvers/activities.py
+++ b/ayon_server/graphql/resolvers/activities.py
@@ -23,7 +23,9 @@ from ayon_server.utils import SQLTool
 
 ACTIVITY_FEED_ALLOWED_KEYS = [
     "activity_type",
+    "activity_data",
     "reference_type",
+    "reference_data",
     "entity_type",
     "entity_name",
     "entity_id",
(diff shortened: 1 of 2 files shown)


martastain (changes requested):
Mostly works, pls just add `activity_data` and `reference_data` to `ACTIVITY_FEED_ALLOWED_KEYS` so it is possible to filter by it as well.

martastain (approved):

PR #934, opened 2026-05-18
User invitations

martastain:
This pull request introduces a new user invitation feature to the API, allowing managers and administrators to invite users via email and enabling invited users to accept the invitation and set their password. The changes include new API endpoints, email template logic, and updates to module exports to support the new functionality.

### DIFF
--- a/api/users/invite.py
+++ b/api/users/invite.py
@@ -109,6 +109,7 @@ async def accept_invite(
 
     user = await UserEntity.load(user_name)
     user.data["inviteRequest"] = None
+    user.data["inviteAcceptedAt"] = time.time()
     user.set_password(payload.password, complexity_check=True)
     await user.save()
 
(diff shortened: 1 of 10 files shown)


api/users/invite.py
```diff
@@ -0,0 +1,121 @@
...
+        "token": token,
+    }
+
+    user.data["inviteRequest"] = invite_request
+    server_url = server_url_from_request(request)
+    server_config = await get_server_config()
+    studio_name = server_config.studio_name
+
+    tplvars = {
+        "reset_url": f"{server_url}/acceptInvite?token={token}",
```
filipvnencak: little typo 
```suggestion
        "reset_link": f"{server_url}/acceptInvite?token={token}",
```

filipvnencak (changes requested):

api/users/invite.py
martastain: ah. sorry. sure

Innders (changes requested):
- [ ] Email says "Reset password" when it should say "Accept invite"

Image

PR #936, opened 2026-05-19
Added allowStagingMode to attributes

### LINKED ISSUE #935: YN-0452: Expose web actions in staging mode

Create a new user flag "allowStagingMode" that will be used to expose a staging toggle for the user (like developerMode).

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1466029580652646461)

_This issue was automatically created from AYON ticket [YN-0452]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=3df8679833f311f196df6a78a2975fec)

kalisp:
## Description of changes

New `allowStagingMode` User attribute added. Currently will be used by FE to control which Web Actions should be triggered.

Not used/exposed in backend in any further way.

### Technical details

`.manage.ps1 setup` must be run on existing installations!

### DIFF
--- a/setup/attributes.py
+++ b/setup/attributes.py
@@ -151,13 +151,6 @@ DEFAULT_ATTRIBUTES: dict[str, dict[str, Any]] = {
         "title": "Developer mode",
         "example": True,
     },
-    "isStagingAllowed": {
-        "scope": "U",
-        "type": "boolean",
-        "title": "Allow staging mode",
-        "example": False,
-        "description": "Allow frontend staging mode to be enabled by this user to access staging web actions.",
-    },
     "productGroup": {
         "scope": "S",
         "type": "string",
--- a/setup/attributes.py
+++ b/setup/attributes.py
@@ -151,7 +151,7 @@ DEFAULT_ATTRIBUTES: dict[str, dict[str, Any]] = {
         "title": "Developer mode",
         "example": True,
     },
-    "allowStagingMode": {
+    "isStagingAllowed ": {
         "scope": "U",
         "type": "boolean",
         "title": "Allow staging mode",
--- a/setup/attributes.py
+++ b/setup/attributes.py
@@ -156,6 +156,7 @@ DEFAULT_ATTRIBUTES: dict[str, dict[str, Any]] = {
         "type": "boolean",
         "title": "Allow staging mode",
         "example": False,
+        "description": "Allow frontend staging mode to be enabled by this user to access staging web actions.",
     },
     "productGroup": {
         "scope": "S",
--- a/setup/attributes.py
+++ b/setup/attributes.py
@@ -151,6 +151,12 @@ DEFAULT_ATTRIBUTES: dict[str, dict[str, Any]] = {
         "title": "Developer mode",
         "example": True,
     },
+    "allowStagingMode": {
+        "scope": "U",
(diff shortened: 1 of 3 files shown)


setup/attributes.py
```diff
@@ -151,6 +151,12 @@
         "title": "Developer mode",
         "example": True,
     },
+    "allowStagingMode": {
+        "scope": "U",
+        "type": "boolean",
+        "title": "Allow staging mode",
+        "example": False,
```
BigRoy: I feel like this attribute may warrant a description? 

"Allow frontend staging mode to be enabled by this user to access staging web actions."

Or whatever is reasonable here? @Innders 
Unless description is something that should be a short description.

Innders:
@martastain suggested it should be `isStaging` like `isDev`, this makes sense to me?

BigRoy:
> @martastain suggested it should be `isStaging` like `isDev`, this makes sense to me?

From a pure code perspective `isDev` makes more sense to me than `isStaging` - just because staging really points to the mode but dev points to being a developer. You'd almost expect it to be `isStager` then but that doesn't sound right. It's as if it were `isDevMode` which would be similarly confusing because just that attribute does not define whether the user is in dev mode or staging mode; it only defines they have access to it so that they _can_ enable it on the frontend.

I understand the reasoning to keep both attributes named in a similar fashion, but I keep thinking it's a confusing naming.
Would `isStagingAllowed` be better?

Innders:
`isStagingAllowed` works for me

martastain:
But there are two different cases. One is for allowing staging mode `data.isStagingAllowed`  - this is set by administrator. Then an attribute, that is user editable (if isStagingAllowed is true) to actually enable it. right? This is how dev mode works.

martastain:
if this is to "allow staging mode" how it is actually turned on?

Innders (approved):

Innders:
> if this is to "allow staging mode" how it is actually turned on?
 
 A switch like dev mode

martastain:
> > if this is to "allow staging mode" how it is actually turned on?
> 
> A switch like dev mode

That's my point. And where that value is stored?

Innders:
> That's my point. And where that value is stored?

Local storage or user preferences?

kalisp:
> > > if this is to "allow staging mode" how it is actually turned on?
> > 
> > 
> > A switch like dev mode
> 
> That's my point. And where that value is stored?

So do I need to update/add anything in this PR to move it forward? (@martastain)

martastain:
I still don't understand why do you insist on an attribute, when all similar flags for user are stored in data.

Innders:
> I still don't understand why do you insist on an attribute, when all similar flags for user are stored in data.

Agree it should go in data

kalisp:
So it actually shouldn't be attribute but flag on `user.data`?

I used `is_staging_enabled` value matching what it in `ayon-python-api`. 

Or what it is use case for this:
- can user enable staging mode -> is_staging_allowed
- has user enabled staging mode -> is_staging_enabled

(But wouldnt be attribute still necessary for user to actually switch some toggle?)

kalisp:
After discussion with Martin renamed back to `is_staging_allowed`.

martastain (approved):

PR #939, opened 2026-05-20
Installer and Dependency package metadata

iLLiCiTiT:
## Description of changes
Added python version and distro short name into installer and dependency package models.
Both Installer and Dependency Package now can have filled `python_path` and `distro_short` which can be used for validation of compatibility.

### Technical details
Moved `python_version` from `InstallerManifest` to `BasePackageModel` which is also used for `DependencyPackageManifest`. With that both can have the value filled, it is optional as old dependency packages don't have it filled. Also added `distro_short` to the manifest as additional information to `platform`.

### Additional context
The goal is to be able to check if installer is compatible with dependency package before it is used in production.

### DIFF
--- a/ayon_server/installer/models.py
+++ b/ayon_server/installer/models.py
@@ -70,6 +70,14 @@ class BasePackageModel(OPModel):
             example="3.11",
         ),
     ] = None
+    distro_short: Annotated[
+        str | None,
+        Field(
+            title="Short distribution name",
+            description="Short name of the Linux distribution",
+            example="rocky9",
+        ),
+    ] = None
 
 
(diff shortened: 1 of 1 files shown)


martastain (approved):
I'll add this to ynputcloud models as well. They need to be in sync

BigRoy:
How would we test whether this data model change is sufficient for our needs and that it works? :) I suppose there's more to do than just the model changes for this to work?

Or is the model change sufficient to take/store it from the JSON already?

iLLiCiTiT:
> Or is the model change sufficient to take/store it from the JSON already?

If it's in the json then you can access it with this PR.

PR #940, opened 2026-05-22
Bugfix: Implemented passing icon/color in project graphQL

### LINKED ISSUE #869: GraphQl: Project productTypes and productBaseTypes fields
## Description
Currently `productTypes` field can return `name`, `icon` and `color`, but `icon` and `color` are always returning `null` value instead of using values from project's product base types config. And `productBaseTypes` don't have `color` and `icon` which probably should have.

kalisp:
## Description of changes

`icon` and `color` of `productTypes` and `productBaseTypes` weren't passed in graphQL query.

This PR fixes this, at least for `project` query.

image

### Technical details

Same code is replicated for `productTypes` and `productBaseTypes` as there is no differentation between them in Settings, eg. Anatomy. I am not sure, if there will be as `productTypes` will be dynamic inherently. Anatomy would need to collect all defined keys from all Settings which seems as a hassle. 

There is also separate `productTypes` which should be removed, imho. This query has description "Studio-wide product type configuration", but there are no 'studio' only `productTypes`/`productBaseTypes` which could store `icon`/`color` information.

There is `public.product_types` table, but it seems it contains only keys of published products (eg. it grows with new type of published product). There is no `data` values currently which could hold information (icon/color) for these. And even if there would be, what should be stored there as `icon`/`color` could be set differently for each of the `productTypes` in each project.
I don't understand purpose of this table.

[GraphQl: Project productTypes and productBaseTypes](https://github.com/ynput/ayon-backend/issues/869)

### DIFF
--- a/ayon_server/graphql/nodes/common.py
+++ b/ayon_server/graphql/nodes/common.py
@@ -16,6 +16,8 @@ from ayon_server.utils import json_dumps
 @strawberry.type
 class ProductType(BaseEdge):
     name: str = strawberry.field()
+    icon: None = strawberry.field(default=None)
+    color: None = strawberry.field(default=None)
 
 
 @strawberry.type
--- a/ayon_server/graphql/nodes/common.py
+++ b/ayon_server/graphql/nodes/common.py
@@ -16,8 +16,6 @@ from ayon_server.utils import json_dumps
 @strawberry.type
 class ProductType(BaseEdge):
     name: str = strawberry.field()
-    icon: str | None = strawberry.field(default=None)
-    color: str | None = strawberry.field(default=None)
 
 
 @strawberry.type
--- a/ayon_server/graphql/nodes/common.py
+++ b/ayon_server/graphql/nodes/common.py
@@ -25,8 +25,8 @@ class ProductType(BaseEdge):
 @strawberry.type
 class ProductBaseType(BaseEdge):
     name: str = strawberry.field()
-    icon: str | None = strawberry.field(default=None)
-    color: str | None = strawberry.field(default=None)
+    icon: str = strawberry.field()
+    color: str = strawberry.field()
 
 
 async def _get_entity(
--- a/ayon_server/graphql/nodes/common.py
+++ b/ayon_server/graphql/nodes/common.py
@@ -23,6 +23,8 @@ class ProductType(BaseEdge):
 @strawberry.type
 class ProductBaseType(BaseEdge):
     name: str = strawberry.field()
+    icon: str | None = strawberry.field(default=None)
+    color: str | None = strawberry.field(default=None)
 
 
 async def _get_entity(
--- a/ayon_server/graphql/nodes/common.py
+++ b/ayon_server/graphql/nodes/common.py
@@ -16,6 +16,8 @@ from ayon_server.utils import json_dumps
 @strawberry.type
 class ProductType(BaseEdge):
     name: str = strawberry.field()
+    # TODO remove 'icon' and 'color' for ayon_api compatibility
+    #     PR https://github.com/ynput/ayon-python-api/pull/332
     icon: None = strawberry.field(default=None)
     color: None = strawberry.field(default=None)
 
--- a/ayon_server/graphql/nodes/project.py
+++ b/ayon_server/graphql/nodes/project.py
@@ -345,17 +345,11 @@ class ProjectNode:
 
     @strawberry.field(description="List of project's product types")
     async def product_types(self) -> list[ProductType]:
-        # TODO separate from productBaseTypes definitions when available
-        # in Anatomy
-        default_color, default_icon, definitions = \
-            await self._get_product_base_type_defs()
         if self.skeleton:
             return []  # TODO: load from skeleton data instead of returning empty list
         return [
             ProductType(
-                name=row["name"],
-                icon=definitions.get(row["name"],{}).get("icon") or default_icon,
-                color=definitions.get(row["name"],{}).get("color") or default_color,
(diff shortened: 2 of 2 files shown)


ayon_server/graphql/nodes/project.py
```diff
@@ -344,13 +345,17 @@ async def link_types(self, active_only: bool = False) -> list[LinkType]:
...
+        default_color, default_icon, definitions = \
+            await self._get_product_base_type_defs()
         if self.skeleton:
             return []  # TODO: load from skeleton data instead of returning empty list
         return [
             ProductType(
                 name=row["name"],
-                icon=row["data"].get("icon"),
-                color=row["data"].get("color"),
+                icon=definitions.get(row["name"],{}).get("icon") or default_icon,
```
iLLiCiTiT: The question, for me, is that the product base types are planned to be expanded to also handle diffirent icons product type per product base type, in that case we won't be able to handle the icon and color with "only" product type.

So, the question is if we should remove the `icon` and `color` to not confuse anyone as it won't be enough?

The issue was created during the product base types refactor and a lot of changed over time.

kalisp: You mean that `productType` doesn't have `icon`|`color` in Settings now, so it shouldn't have these fields and not be re-using values for `productBaseType`? (eg. format of models will flip, `ProductBaseType` will gain `icon`|`color`, `ProductType` will loose them.

That makes sense to me. (I'll wait for Martin response.)

iLLiCiTiT: No, the logic is that we do define icon & color per "product base type", with option to change that per "product type".

Like this
```
- render (cat #ffffff)
    - bg (dog #ff0000)
- image (image #00ff00)
    - bg (tree #0000ff)
```
`render` and `image` are product base types, and `bg` is product type. As you can see `bg` from render have different icon and color than `bg` from `image`.

kalisp: So should this PR add separate `product_types` (and rename existing label for `product_base_types`) in `https://github.com/ynput/ayon-backend/blob/develop/ayon_server/settings/anatomy/__init__.py` and actually pull from there?

But again, arent `product_types` dynamic, completely non standardized? Would be responsibility of admin who would like to have specific icon/color to add any newly defined `product_type`? (So defaults would be completely empty. Or actually list of `product_types` would be empty, there might be same `default` as it is for `product_base_types` in case, you would like to just differentiate between `real` (base) type and `label` (`product_type`).

iLLiCiTiT: > So should this PR add separate product_types (and rename existing label for product_base_types) in https://github.com/ynput/ayon-backend/blob/develop/ayon_server/settings/anatomy/__init__.py

Considering the source issue, no, don't change that. That would require to enhance the model of the product base types, to allow specify product types per product base type, which would also require to change frontend, is probably out of scope of the issue.

> and actually pull from there?

That is my point, you can't pull anything from there, you can't get icon and color for product type if you have only product type. To be able to get icon and color for product type you must have both product base type and product type. The issue is describing that right now GraphQl does offer `icon` and `color` for `productTypes` field, but there is no way how server can fill them -> the values are either misleading or empty. On the other side `productBaseTypes` field does offer only `name` but could offer `icon` and `color`.

> Would be responsibility of admin who would like to have specific icon/color to add any newly defined product_type?

Yes, that's how the anatomy product base types are meant to be. That's why there is "default" in case none of the cases mentioned in the other definitions do match.

image

BTW Description does not mention the issue https://github.com/ynput/ayon-backend/issues/869

ayon_server/graphql/nodes/common.py
```diff
@@ -16,13 +16,13 @@
 @strawberry.type
 class ProductType(BaseEdge):
     name: str = strawberry.field()
```
iLLiCiTiT, line 18: I'm afraid we can't remove it. It is being used in ayon api, so removing it would crash. But not filling it is good start.
```suggestion
    name: str = strawberry.field()
    icon: None = strawberry.field(default=None)
    color: None = strawberry.field(default=None)
```

ayon_server/graphql/nodes/common.py
```diff
@@ -16,13 +16,15 @@
...
-    icon: str | None = strawberry.field(default=None)
-    color: str | None = strawberry.field(default=None)
+    icon: None = strawberry.field(default=None)
```
iLLiCiTiT, line 21: https://github.com/ynput/ayon-python-api/pull/332
```suggestion
    # TODO remove 'icon' and 'color' for ayon_api compatibility
    #     PR https://github.com/ynput/ayon-python-api/pull/332
    icon: None = strawberry.field(default=None)
```

iLLiCiTiT (approved):

ayon_server/graphql/nodes/common.py
```diff
@@ -16,13 +16,17 @@
...
+    #     PR https://github.com/ynput/ayon-python-api/pull/332
+    icon: None = strawberry.field(default=None)
+    color: None = strawberry.field(default=None)
 
 
 @strawberry.type
 class ProductBaseType(BaseEdge):
     name: str = strawberry.field()
+    icon: str | None = strawberry.field(default=None)
+    color: str | None = strawberry.field(default=None)
```
iLLiCiTiT:
```suggestion
    icon: str = strawberry.field()
    color: str = strawberry.field()
```

martastain:
The product_types table is there now only for backwards compatibility. We cannot get rid of it because it would break project imports.

martastain (approved):

PR #941, opened 2026-05-25
Skip failing only actions

### LINKED ISSUE #920: Actions: Failed collection
## Story

If one addon has bug in action manifest collections it prevents to collect rest of actions from other addons.

### To reproduce

1. Override `get_simple_actions` in an addon and raise an error.
2. No actions will be displayed at that moment.

### Expected behavior

It should log the error and skip actions of the addon, but should show actions of other addons.

## Environment

- **Server version**: 
- **Server host OS**: 
- **Browser**: 
- **Client version**:

kalisp:
## Description of changes

This PR is catching errors in collection of addons `actions`. Possible errors are logged and addon action collection is skipped.

Without this failure in any addon limited showing up all other actions from other addons.

### DIFF
--- a/ayon_server/actions/listing.py
+++ b/ayon_server/actions/listing.py
@@ -10,7 +10,7 @@ from ayon_server.events import EventModel, EventStream
 from ayon_server.exceptions import NotFoundException
 from ayon_server.lib.postgres import Postgres
 from ayon_server.lib.redis import Redis
-from ayon_server.logging import logger
+from ayon_server.logging import logger, log_traceback
 from ayon_server.types import Field, OPModel
 from ayon_server.utils import json_dumps, json_loads
 
@@ -262,7 +262,7 @@ class SimpleActionCache:
                 cached_data = [x.dict() for x in r]
                 result = [SimpleActionManifest(**x) for x in cached_data]
             except Exception as e:
-                logger.error(
+                log_traceback(
                     "Failed to get simple actions for addon "
                     f"{addon.name} v{addon.version}: {e}"
                 )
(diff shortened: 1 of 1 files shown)


ayon_server/actions/listing.py
```diff
@@ -256,12 +256,22 @@ async def get(
...
-            cached_data = [x.dict() for x in r]
+            try:
+                r = await addon.get_simple_actions(project_name, variant)
+                # Cache the data
+                cached_data = [x.dict() for x in r]
+                result = [SimpleActionManifest(**x) for x in cached_data]
+            except Exception as e:
+                logger.error(
+                    "Failed to get simple actions for addon "
+                    f"{addon.name} v{addon.version}: {e}"
```
BigRoy, line 267: Should this pass `exc_info=True` to the logger just so we get access to the full stack trace in the log?

ayon_server/actions/listing.py
```diff
@@ -333,7 +343,16 @@ async def get_dynamic_actions(
...
     variant, addons = await get_relevant_addons(variant, user)
     for addon in addons:
-        for action in await addon.get_dynamic_actions(context, variant):
+        try:
+            dynamic_actions = await addon.get_dynamic_actions(context, variant)
+        except Exception as e:
+            logger.error(
+                "Failed to get dynamic actions for addon "
+                f"{addon.name} v{addon.version}: {e}"
+            )
```
BigRoy: Should this pass `exc_info=True` to the logger just so we get access to the full stack trace in the log?

martastain: You may use 

```python
log_traceback(
    "Failed to get dynamic actions for addon {addon.name} {addon.version}"
)
``` 

from `ayon_server.logging` module in the except block - it formats the stack trace and stores it in the event stream properly

martastain: image

kalisp: Ook

martastain (approved):

PR #943, opened 2026-05-26
Added calculation of statistical info for column values

### LINKED ISSUE #937: YN-0755: Overview column summaries

We need to provide option to show a summary of all the values in each column in overview and products pages.  

### Additional context:

_This issue was automatically created from AYON ticket [YN-0755]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=1b8d87ec082848e99cd918dfbce496f2)

kalisp:
## Description of changes

WIP

Adds list of statistical information about column values. Will be used for FE to display how many records have column filled/empty, min/max values for integer values etc.

Implemented for:
- folders
-products
- versions
- tasks
- entitylistitems

image

image

This PR is just a starting point for discussion, having issues with potential expensiveness of duplicated queries, didn't test pagination properly.


martastain (changes requested):
This is not scalable. You'd potentially load all version data from the database only to do a calculation. That will break stuff. aggregation need to be done sql-side, preferably the resolver should return either the result or pagination, but not both at the same time - that way, you can just change the behavior by changing sql_columns. Returning aggregations and actual resultset at the same time does not make much sense, since you only load statistics once when filter changes, but the resultset is paginated.

kalisp:
Reworked to build sql approach and either use it (via `get_folders.calculate_statistics`) or stick with original logic.

Currently it calculate for hand selected list of columns:

image

  as I am guessing only specific columns might be of interest (and calculate it for everything seems wrong.)
  
  But if necessary, I would need to dynamically get list of attrib/project_attributes/inherited_attributes and use them.
  
I am bit worried how FE will match returned names (based on column names/labels) to edge keys. (Open for suggestions.)

(Still WIP.)

martastain (changes requested):
Could you please refactor the PR, so classes and helper functions are in a separate file? Probably somthing like `ayon_server/graphql/resolver/field_stats.py`? 
The plan is to call all the fields (attrib and top level fields) "fields" eventually so that would make sense. Then you could have:

```python

def generate_field_stats(query: str) -> list[ColumnStats]:
    ...
```

That you could call directly from the resolver:

```python

if calculate_statistics:
    field_stats = generate_field_stats(query)

    return FoldersConnection(
        edges=[]
        page_info=PageInfo(
            has_next_page=False,
            has_previous_page=False,
            start_cursor=None,
            end_cursor=None,
            column_metadata=field_stats
        )
      )

# or else

return await resolve(
    FoldersConection,
    FolderEdge,
    FolderNode,
    query,
    ...
)

```


As for the naming and field locations:

I'd maybe use `field_stats` everywhere and put it to the same level as PageInfo, not as a part of it (but i am willing to be convinced otherwise). 
The reason is that field stats are not really a part of page info, they are more like metadata about the fields in the query. 
So it would make more sense to have it as a separate field in the response rather than nesting it under page info.


And the most important bit:
Is this supposed to work? The query generation seems to be broken

Image

Image

kalisp:
> And the most important bit:
> Is this supposed to work? The query generation seems to be broken

I was testing it bit differently, didn't notice that pure graphQl is not providing `has_reviewables` CTE.

filipvnencak (changes requested):
Missing enums and breakdown for each enum aslo assignee it shold behave same as Enums
Something like:
```
{ "columnName": "status",
    "distribution": [
      { "value": "in_progress", "count": 120 },
      { "value": "done", "count": 340 }
       ....
    ] } 
```

For numbers add **sum** too, there already is min/max/avg

and also counting for tasks, versions products now there is only for folders

kalisp:
> Missing enums and breakdown for each enum aslo assignee it shold behave same as Enums Something like:
> 

Implemented `SUM` and `DISTRIBUTION`.
Screenshot 2026-06-04 111237

filipvnencak (changes requested):
Assignees and tags still missing, also Lists page support is still missing

kalisp:
Distribution and counts should be working on `tags` (and similar).

Implemented calculations on `get_entity_list_items`, not sure if this one is what you are looking at.

martastain (approved):
lgtm

PR #946, opened 2026-06-01
Unified folder and task search

martastain:
This pull request introduces a new unified text search feature to the folders API, allowing users to perform a global search across both tasks and folders. The main changes include adding a new `search` parameter to the request model and updating the SQL query logic to support searching by multiple terms across relevant fields.

**Global Search Feature:**

* Added a `search` field to the `FolderSearchRequest` model, allowing clients to specify a global text search term that will be applied to both tasks and folders (`api/folders/search.py`).

### DIFF
--- a/api/folders/search.py
+++ b/api/folders/search.py
@@ -40,6 +40,14 @@ class FolderSearchRequest(OPModel):
         ),
     ] = None
 
+    search: Annotated[
+        str | None,
+        Field(
+            title="Global Text search",
+            description="Unified 'fulltext' search applied to both tasks and folders",
+        ),
+    ] = None
+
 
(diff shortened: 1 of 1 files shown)


filipvnencak (approved):

Innders (changes requested):
I can't seem to get it to work :/

Image

Image

martastain:
You need to call it with `search` argument. not `folderSearch` and `taskSearch`. 

`folderSearch` and `taskSearch` work as before - they basically use `AND`, because they affect different subqueries, whereas `search` filters  tasks and folders at the same time, so it (should) mimic `OR` operator

Innders (approved):

PR #963, opened 2026-06-09
Implemented status filter for resolve api

### LINKED ISSUE #944: YN-0772: USD Resolver: Support latest approved version

The resolve endpoint on the backend currently supports 'latest' but does not support 'latestApproved'. Add support to also be able to resolve 'latestApproved'.

An alternative would be to support:
- version=latest&status=approved
 

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1494321623606628526/1501147845674143806)

_This issue was automatically created from AYON ticket [YN-0772]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=3b0b44e6c1a94a9fbd90fa00201b8177)

kalisp:
## Description of changes

This PR implements additional filtering by statuses. It could be used to find `latest AND Approved` version for example.

Also allows to filter on `Status.state` (`in_progress` == ['In progress', 'Pending review'])

### Technical details

It supports filtering by multiple (OR) statuses:

`"ayon+entity://ayon_project/shots/sh01?version=latest&status=In progress&status=Approved"`
`"ayon+entity://ayon_project/shots/sh01?version=latest&status_state=done&status=Blocked"`


BigRoy (commented):
Although this `statuses` query is explicit I'm not sure it matches with other aspects of AYON statuses.

Statuses in AYON have a "state"
Image

And most likely what you'd want by default is to filter by anything that's at the "Done" state instead of requiring the entity URI to be status name specific. 

To my knowledge on the Products tab when you sort it by featuring **Latest done** it's essentially prioritizing those that have the 'Done' state - regardless of what status it is.

Image

I feel like with the resolver it's likely you'll want that.

I think the `statuses` isn't useless - I feel like we just need some way to also accept `state` as an input perhaps. Then "latest approved' as we'd query it would most likely turn out to be `version=latest&state=done` but if someone wants to get very explicit about getting a particular status, they could `version=latest&status=Approved`.

I do have concerns about the keys `state` and `status` being so darn close to each other. 🗡️ 

Note that `state=Done` is essentially equivalent to `status=X&status=Y&status=Z` for all statuses of that state. However, USD entity URIs will be embedded in tons of scenes, and just renaming a status name would essentially kill existing ones if you were using the explicit status names, etc. I feel like using `state` for that helps to guard against that some more.

kalisp:
I agree that filtering by `state` might be useful, should that be in this PR too?

BigRoy:
> I agree that filtering by `state` might be useful, should that be in this PR too?

I think it's the more important one of the two and may fit more with the ticket/request itself actually so yes, let's keep in here. Having `status=` supported is then just a bonus because you'd already done that work. 😄

kalisp:
Implemented filtering by `status_state`.

martastain (approved):
the code looks good. not a big fan of `status_states` name, but that's just me :) wouldn't `states` be enough?

BigRoy:
> the code looks good. not a big fan of `status_states` name, but that's just me :) wouldn't `states` be enough?

From the URI/query itself I'd also prefer `state` over `status_state`.

kalisp:
> From the URI/query itself I'd also prefer `state` over `status_state`.

Renamed to `state`

api/resolve/models.py
```diff
@@ -105,3 +105,14 @@ class ParsedURIModel(OPModel):
...
+        None,
+        title="Statuses",
+        description="List of statuses to filter by",
+    )
+    states: list[str] | None = Field(
+        None,
+        title="Status states",
+        description="List of status states used for filtering",
+        example=["in_progress", "done"],
+    )
```
BigRoy:
```suggestion
    )

```

BigRoy:
As a side-question. How trivial would it be to add tests for these endpoints @martastain @kalisp ?

mkolar:
I'm late to the game, but I agree that state is way more important at the moment than status here

BigRoy (changes requested):
@kalisp Thanks - I kind of missed that the status/state would apply to the resolved entity, so taking either status from the product or version or representation. Sorry about that.

After a chat with Milan we agreed that we're overcomplicating things, it's really about the version status and we should match what the frontend does with its filters, which is allowing the `version` to be `latest`, `hero` or `latestDone`.

In short, the thing we _actually_ need to do is to remove `state` and `status` query arguments,
And just support `version=latestDone`. Mixing `version` with `status` and `state` and what not will just be overcomplicating things, and the frontend doesn't do that for a reason.

kalisp:
Closing this PR as it got FUBARed because of reverts of reverts. Will be replaced by clean branch.

PR #964, opened 2026-06-09
Version review activities

jakubfiala:
- added a new human activity type 'version.review'. This will be used in review sessions.
- added a `mine` argument to the `get_activities` resolver which lets us filter for activities of current user

### Technical details

related to https://github.com/ynput/ayon-review/issues/213

### DIFF
--- a/api/activities/activity.py
+++ b/api/activities/activity.py
@@ -32,7 +32,7 @@ from ayon_server.types import Field, OPModel
 from .router import router
 
 
-human_activity_types = ['comment', 'guest_review']
+human_activity_types = ['comment', 'version.review']
 
 async def delete_unused_files(project_name: str) -> None:
     storage = await Storages.project(project_name)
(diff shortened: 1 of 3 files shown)


martastain (changes requested):
I'm not a big fan of `guest_review` - i'd use something more generic, such as `version.review` or similar (we already use `version.*` prefix for publishing)

jakubfiala:
@martastain I renamed the activity type as you suggested. I thought I'd also add the other change needed for version review activities - basically there's a requirement to find the last activity of type `version.review` authored by the current user, so I added a `mine` boolean arg to the resolver.

I thought I'd be defensive and not make it a full on `userName` argument, so that a given user cannot enumerate all activities for any other user. However, if you'd prefer to have it that way, let me know.

martastain (approved):

Innders:
Will `version.review` activities trigger inbox messages (notifications). You will want to know that some left a review and I would class it as important.

PR #974, opened 2026-06-11
Removed strongly typed View Settings

### LINKED ISSUE #971: Open up all views
Currently some views like the overview, taskProgress, products and lists have strictly typed settings models.

This is quite annoying as any frontend changes also require backend changes just to add one small field.

Remove these strictly typed settings so that we can store any data in settings.

kalisp:
## Description of changes

This PR removes strongly typed View Settings for some tabs as it was too cumbersome for frontend to be using them. Any change in FE required same change in BE.

This PR should simplify development.


martastain:
I agree, but the problem is that strong types (using OPModel) do camelCase -> snake_case conversion by default. I am afraid that by removing the types, already stored views will break

kalisp:
> I agree, but the problem is that strong types (using OPModel) do camelCase -> snake_case conversion by default. I am afraid that by removing the types, already stored views will break

So should there be some one-time backend DB conversion script to resave it in `camelCase` or should FE be handling it?

Innders (changes requested):
This seems to have broken patching of views

`[PATCH] http://localhost:3000/api/views/versions/3956a018cf9d4f64b44723bf673bc2d5?project_name=wing_it`

```json
{
    "settings": {
        "filter": {
            "operator": "and",
            "conditions": []
        },
        "columns": [],
        "sort_by": null,
        "group_by": null,
        "show_grid": false,
        "sort_desc": false,
        "row_height": 34,
        "grid_height": null,
        "slicer_type": "hierarchy",
        "show_products": false,
        "show_empty_groups": false,
        "group_sort_by_desc": false,
        "featured_version_order": null,
        "showEmptyGroups": false,
        "rowHeight": 34
    }
}
```

```json
{
    "code": 400,
    "detail": "Request validation error in [PATCH] /api/views/versions/3956a018cf9d4f64b44723bf673bc2d5",
    "path": "/api/views/versions/3956a018cf9d4f64b44723bf673bc2d5",
    "traceback": "viewType: field required",
    "errors": [
        {
            "loc": [
                "body",
                "viewType"
            ],
            "msg": "field required",
            "type": "value_error.missing"
        }
    ]
}
```

martastain:
first check if i'm right, but i think i am. then i thing the best place to do that would be here: https://github.com/ynput/ayon-backend/blob/enhancement/open_view_settings/api/views/models.py#L130

keep the existing data in the database and upon returning, check for well known keys in the json. for exampe if view type is overview and there's "sort_by", just convert to camel case. upon next save, it will be overwritten with camelCase variant and reprocessing won't be triggered again.

martastain:
this should bump minor version, since after updating, downgrading would cause a data loss @Innders

kalisp:
> This seems to have broken patching of views

Isn't that issue of FE though? I am now sending completely new model/content, so FE should take that into account, no? Or am I just dense (highly possible).

image

Innders:
> Isn't that issue of FE though? I am now sending completely new model/content, so FE should take that into account, no? Or am I just dense (highly possible).

Nothing has changed in the FE and I'm not sure why it should change?

kalisp:
Added optionality to `view_type` which got rid of the issues with validations.

Innders (changes requested):
Before this change when we generated the queries the mutations would use `payload`.

Now they use `genericViewPostModel` and ``genericViewPatchModel`.

```
export type UpdateViewApiArg = {
  viewType: string
  viewId: string
  projectName?: string
  genericViewPatchModel: GenericViewPatchModel
}
```

can we please figure out a way so that the this goes back to `payload`

```
export type UpdateViewApiArg = {
  viewType: string
  viewId: string
  projectName?: string
  payload: GenericViewPatchModel
}
```

Tagging @filipvnencak

api/views/views.py
filipvnencak:
```suggestion
    payload: Annotated[ViewPostModel, Body(title="Payload")],
```

filipvnencak:
```suggestion
    payload: Annotated[ViewPatchModel, Body(title="Payload")],
```

api/views/models.py
```diff
@@ -203,38 +77,23 @@ class GenericViewPatchModel(BaseViewPatchModel):
...
-    | ListsViewPostModel
-    | ReviewsViewPostModel
-    | VersionsViewPostModel
-    | GenericViewPostModel,
+    GenericViewPostModel,
     Field(
         discriminator="view_type",
         title="View post model",
     ),
 ]
```
filipvnencak:
```suggestion
ViewPostModel = GenericViewPostModel
```

api/views/models.py
filipvnencak:
```suggestion
ViewPatchModel = GenericViewPatchModel
```

filipvnencak (changes requested):
With these backend tweaks the OpenAPI body comes through as payload again, so the frontend doesn't need the codegen-rename workaround in ynput/ayon-frontend#2066

api/views/views.py
filipvnencak:
```suggestion
from fastapi import Path, Query, Request, Body
```

filipvnencak (changes requested):

filipvnencak (approved):

kalisp:
> this should bump minor version, since after updating, downgrading would cause a data loss @Innders

I added tag, but I am not sure if anything actually changed, format wise. I tried to set up bunch of Views and do an upgrade and content seemed same as before.

martastain:
I just reproduced the problem i was expecting. It affects fields in previously strictly typed settings that contain underscores, such as overview.show_hierarchy, overview.sort_by etc.

Steps to reproduce:

- in develop, create an overview view that sets group_by and  row_height 
- switch to this branch
- load the view
- group_by and row_height is not preserved.

it only does not affect filter and columns, as they don't use underscores.

possible solution:

update `construct_view_model` function to something like this.

```python
def construct_view_model(**data: Any) -> ViewModel:
    view_type = data["viewType"]
    settings = data["settings"]
    if view_type in ["overview", "taskProgress", "lists", "reviews", "versions"]:
        patched_settings = {}
        for key, val in settings.items()
            if "_" in key:
                patched_settings[camelize(key)] = val
            else:
                patched_settings[key] = val
        settings = patched_settings
    return GenericViewModel(view_type=view_type, settings=settings)
```

kalisp:
> I just reproduced the problem i was expecting

F00k, didn't notice it. Implemented your comment. Added TODO to remove it in the future.

image

martastain:
>  Added TODO to remove it in the future.

Nothing is as permanent as a temporary workaround :-D

martastain (approved):
Works. I love when we remove code :)

PR #975, opened 2026-06-11
Add 'authors' argument to get_activities GraphQL resolver

jakubfiala:
## PR Checklist

## Description of changes

Adds the `authors` argument to the GetActivities resolver.

### Additional context

related to https://github.com/ynput/ayon-review/issues/213

### DIFF
--- a/ayon_server/graphql/resolvers/activities.py
+++ b/ayon_server/graphql/resolvers/activities.py
@@ -74,6 +74,9 @@ async def get_activities(
     changed_after: Annotated[
         str | None, argdesc("Filter activities updated after this timestamp")
     ] = None,
+    mine: Annotated[
+        bool, argdesc("Filter for only the current user's activities")
+    ] = False,
     filter: Annotated[
         str | None, argdesc("Filter activities using QueryFilter")
     ] = None,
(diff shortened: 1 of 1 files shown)


BigRoy:
Isn't this better solved with a filter on `author` and you just pass the username to the query itself?

Or if we want to expose `author` - to then just have a `author` argument you can pass where you pass it the username.
The `mine` seems overly hardcoded?

martastain:
This is a good point. Honestly - i'm not sure (but i tend to agree with Roy here). In both cases it is a shortcut, because you can already do both using filter. So i feel this is more a convenience helper for debugging (because writing filters by hand is annoying). We have similar shortuts across the most of the resolvers and they usually accept arrays like `authors: ["nebuchadnezzar"]` that resolves to:

```
activity_data->>'author' = ANY({SQLTool.array(authors, curly=True)})
```

jakubfiala:
@martastain @BigRoy I brought this up a couple of times in the previous PR and in Discord. The idea was to prevent a user from fetching literally all activities of another user. However, if it's already possible using the filter, then there's no need to be this defensive. I'll change this to the array field as Martin proposed.

jakubfiala:
@martastain updated!

martastain (approved):
Works

Image

PR #977, opened 2026-06-12
Implemented check of installer Python version vs dependency package

### LINKED ISSUE #938: Bundles: Compatibility check Python version between Launcher and dependency package
## Story

For a bundle we will need to perform a compatibility check on the server to validate the Python version of a dependency package <-> launcher. With the move from Py3.9 to Py3.11 on the client desktop side it'd be helpful to see if they are compatible by Python version. 

Both launcher and dependency package manifests will start to contain the the Python version metadata. 

- Launcher jsons have e.g. "python_version": "3.9.13".
- The dependency packages are still lacking this data. If it's lacking it, we should assume is 3.9 and (allow matching against any 3.9.x). It'll start getting this data since this PR: https://github.com/ynput/ayon-dependencies-tool/pull/47

### Problems

We need to perform the compatibility check - for the frontend to show it's incompatible on the bundles page.

### Proposal

Add that compatibility check, and make sure it shows on the frontend

## TODO

- [x] Add relevant data to the backend model: https://github.com/ynput/ayon-backend/pull/939
- [ ] Add compatibility check between the dependency package and launcher manifests

### Attachments

[src discord thread](https://discord.com/channels/517362899170230292/1506218711793532968)

kalisp:
## Description of changes

As Python version changes in recent installers (launchers) this adds check for compatibility of launcher and dependency package set on a bundle.

(Could be tested by triggering `api/bundles/check` - requires BundleModel in body though. That could get pulled via `api/bundles` endpoint.)

image

### DIFF
--- a/api/bundles/check_bundle.py
+++ b/api/bundles/check_bundle.py
@@ -6,7 +6,7 @@ from api.desktop.deps import get_manifest
 from api.desktop.installers import get_installers
 from api.desktop.installers import get_manifest as get_installer_manifest
 from ayon_server.addons.library import AddonLibrary
-from ayon_server.exceptions import NotFoundException
+from ayon_server.exceptions import AyonException, NotFoundException
 from ayon_server.types import Field, OPModel
 from ayon_server.version import __version__ as ayon_version
 
@@ -201,13 +201,17 @@ async def check_bundle(
                 )
 
     # Check for Python version installer compatibility with dependency packs
-    if bundle.dependency_packages:
-        variant: Literal["production", "staging"] | None = (
-            "production" if bundle.is_production
-            else "staging" if bundle.is_staging
-            else None
+    if not bundle.dependency_packages:
+        issues.append(
+            BundleIssueModel(
+                severity="error",
+                addon=None,
+                message=(f"Bundle {bundle.name} doesn't have dependency packages."),
+                required_addon=None,
+            )
         )
-
+    else:
+        # check pythons in all packages
         for platform_name, package_filename in bundle.dependency_packages.items():
             if package_filename is None:
                 continue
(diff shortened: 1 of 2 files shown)


api/bundles/check_bundle.py
```diff
@@ -197,5 +200,65 @@ async def check_bundle(
...
+                    f"uses {installer_python_version} on platform '{platform_name}'."
+                )
+                issues.append(
+                    BundleIssueModel(
+                        severity="error",
+                        addon=None,
+                        message=msg,
+                        required_addon=None,
+                    )
+                )
```
iLLiCiTiT:
```suggestion
            for installer in installer_list.installers:
                installer_manifest = get_installer_manifest(installer.filename)
                installer_python_version = installer_manifest.python_version
                if not installer_python_version:
                    continue
            
                if is_compatible(
                    package_python_version, installer_python_version
                ):
                    continue

                msg = (
                    f"Dependency package '{package_filename}' requires Python"
                    f" {package_python_version}, but installer"
                    f" '{installer.filename}' uses {installer_python_version}"
                    f" on platform '{platform_name}'."
                )
                issues.append(
                    BundleIssueModel(
                        severity="error",
                        addon=None,
                        message=msg,
                        required_addon=None,
                    )
                )
```

api/desktop/installers.py
```diff
@@ -263,3 +229,47 @@ async def patch_installer(user: CurrentUser, filename: str, payload: SourcesPatc
...
+
+
+async def get_installers(
+    version: str | None = None,
+    platform: Platform | None = None,
+    variant: Literal["production", "staging"] | None = None,
+) -> InstallerListModel:
+    result: list[Installer] = []
+
+    if variant in ["production", "staging"]:
```
iLLiCiTiT, line 239: Looks like dev bundles are not handled? Well, probably are because the logic didn't change, but I wonder how? 🙂

BigRoy: @kalisp @martastain any idea here?

kalisp: I think that `variant` is just shortcut when you don't want/do not know exact 'version'.

It is just 'Get me installers for 'production'/'staging' for 'windows''. You cannot ask the same for dev bundles as there could be multiple.

PR #980, opened 2026-06-16
Removed user restriction for link deletion

### LINKED ISSUE #955: YN-0808: Permissions: User should be able to remove links created by other users

Entity links generated by others can't be removed or changed by a "User".

Open the permissions so that by default anyone can remove links.

In later iterations permissions can be added, but not now.

### Additional context:

_This issue was automatically created from AYON ticket [YN-0808]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=65aee5e0631511f18a17bd45f163a507)

### Previous note

_We'll need to expose this to Permissions model so we can start to allow it._

_Side note:_

- _If we were to implement it the same way as other permissions, after update, artist will have write access to links, so we should warn the admins to update their access groups_

kalisp:
## Description of changes

It was decided that everybody should be able to delete link for now.

### Additional context

It might be needed to add proper ACL in the future, but for now simple removal of user check is enough.

Closing [YN-0808](https://github.com/ynput/ayon-backend/issues/955)

### DIFF
--- a/api/links/links.py
+++ b/api/links/links.py
@@ -378,8 +378,11 @@ async def _check_access(
     if not user.is_manager:
         input_class = get_entity_class(input_type)
         input_entity = await input_class.load(project_name, input_id)
-        await input_entity.ensure_update_access(user)
 
         output_class = get_entity_class(output_type)
         output_entity = await output_class.load(project_name, output_id)
-        await output_entity.ensure_update_access(user)
+
+        try:
+            await input_entity.ensure_update_access(user)
+        except ForbiddenException:
+            await output_entity.ensure_update_access(user)
--- a/api/links/links.py
+++ b/api/links/links.py
@@ -379,10 +379,10 @@ async def _check_access(
         input_class = get_entity_class(input_type)
         input_entity = await input_class.load(project_name, input_id)
 
-        output_class = get_entity_class(output_type)
-        output_entity = await output_class.load(project_name, output_id)
-
         try:
             await input_entity.ensure_update_access(user)
         except ForbiddenException:
+            output_class = get_entity_class(output_type)
+            output_entity = await output_class.load(project_name, output_id)
+
             await output_entity.ensure_update_access(user)
--- a/api/links/links.py
+++ b/api/links/links.py
@@ -332,9 +332,6 @@ async def delete_entity_link(
         link_type = res["link_type"]
         link_type_name, input_type, output_type = link_type.split("|")
 
-        if res["author"] != user.name and not user.is_manager:
-            raise ForbiddenException("You do not have permission to delete this link.")
-
         query = "DELETE FROM links WHERE id = $1"
         await Postgres.execute(query, link_id)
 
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Seems to work for me now.

martastain:
This means really "everybody" as long you have ANY permission to the project, you can delete a link. even if you are otherwise fully read only. is that really what we want? shouldn't it at least do something like this?

```python
from ayon_server_helpers.get_entity_class import get_entity_class

if not user.is_manager:
    input_class = get_entity_class(input_type)
    input_entity = await input_class.load(project_name, input_id)
    await input_entity.ensure_update_access(user)

    ouput_class = get_entity_class(output_type)
    output_entity = await ouptut_class.load(project_name, output_id)
    await output_entity.ensure_update_access(user)
```

or maybe ensure you have update access to at least one of the entities?

BigRoy:
> This means really "everybody" as long you have ANY permission to the project, you can delete a link. even if you are otherwise fully read only. is that really what we want? shouldn't it at least do something like this?

I guess being able to remove the link should be based on the fact that you at least are able to access one of the two entities, yes. So that does make more sense.

kalisp:
Added recommended access check.

kalisp:
Actually I am checking for access for both entities, should I check only for `input` one?

BigRoy:
> Actually I am checking for access for both entities, should I check only for `input` one?

I feel like at least one of the too should be allowed access - so if the first is allowed you don't need to check the other?

kalisp:
> I feel like at least one of the too should be allowed access

Modified to check to at least one is accessible.

api/links/links.py
```diff
@@ -357,3 +364,25 @@ async def delete_entity_link(
...
+    input_id: str,
+    output_id: str
+):
+    """Checks that user is manager or has access to both sides of link."""
+    if not user.is_manager:
+        input_class = get_entity_class(input_type)
+        input_entity = await input_class.load(project_name, input_id)
+
+        output_class = get_entity_class(output_type)
+        output_entity = await output_class.load(project_name, output_id)
```
BigRoy: Would putting this in the `except` be able to speed this up ever so slightly if the inputs are accessable? If so, I wonder if it's worth putting it in the except.

Or even structure as:
```
try:
    await input
    return
except ForbiddenException:
    # Check if output entity at least is accessible
    pass

output_class = get_entity_class(output_type)
...
await output
```

kalisp: Moved to except part to resolve only if totally necessary.

BigRoy (approved):
Works for me.

Note that through frontend this can't be deleted because it's "restricted" in the UI itself if you have access to one of the two entities - even though technically the API now allows to delete it as soon as you have at least visibility to one of the entities.

image

Links can be deleted fine though if user has access to both the entity ids from the frontend.

Anyway, from API:
- Can't delete if user has no access to both entities
- Can delete if it has access to at least one of the entities (if you have the link id that is)

But in practice they could reasonably only delete them if they have access to both because frontend doesn't allow it otherwise. Which is fine I'd say.

martastain (approved):

PR #992, opened 2026-06-22
Added activity when attribute gets changed

### LINKED ISSUE #984: YN-0316: Entity changes show as activities in feed
## Background Context
The history of changes for custom attributes or perhaps for any defined set of attributes would be benefitial because some attributes are changed during production and it is good to track it changes similar style as blames. 

Either it could be simply added to the Activty and filtrable somehow so it could be easy to search for. Or we could enhance frontend on individual attribute so with some action a user could look on changes history.

Particualar use case:

Version Entity Clip IN/OUT for plate products - so supervisors could monitor edititng changes

Shot Folder Entitiy - Frame Start/End - comp artist or team leads could monitor when and who had changed it

FPS, Width, Height, PixelAspect attributes likewise. 

## Task

- Create a new activity type for field changes.
- This works the same way as status and assignee changes but all other custom attributes.
- Folder/task/product type should be included

## Additional context:

_This issue was automatically created from AYON ticket [YN-0316]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=1d64dde66a5211f18d276e0e842d8c7d)

kalisp:
## Description of changes

This PR adds activity creation after attribute gets changed to log it of posterity.

This should be used in production to backtrack changes to attributes as `fps`, `width`, `height`


kalisp:
Probably outside of this PR but I am not understanding this part of existing code:

image

kalisp:
Not completely sure, if this is actually solving requested use case. I followed existing event trigger in `ayon-backend/api/tasks/tasks.py` for change of assignee, but maybe I should have use `ayon-backend/ayon_server/activities/event_hook.py`

kalisp:
I think I misunderstood the use case. I reverted commits which triggered event IF attribute definition was changed.

I implemented creating of activity when attribute changes, in similar fashion as `status` change.

But there will be need to update frontend a bit as right this changes is breaking display of entity with newly created activity.

I'll try to experiment with FE for a bit if I could figure it out myself, but if not, I'll punt it FE team way.

martastain:
I am REALLY concerned about this running when a project uses calculated attributes tbh.

kalisp:
You mean it will be spawning a too many activities? Can we separate attribute changes of calculated attributes?

martastain:
Maybe. It will require a changes both in calculated attributes and the server. I'll take a look how that could be done. I don't assume it is necessary to track changes of calc attrs anyways.

Edit: This is now available

ayon_server/activities/event_hook.py
```diff
@@ -75,6 +79,30 @@ async def handle_status_changed(cls, event: "EventModel"):
...
+        assert event.project is not None, "Project is required for activities"
+        entity = await entity_class.load(event.project, event.summary["entityId"])
+
+        old_value = event.payload.get("oldValue")
+        new_value = event.payload.get("newValue")
+
+        origin_link = f"[{entity.name}]({entity_type}:{entity.id})"
+        body = f"{origin_link} attrib changed from {old_value} to {new_value}"
+
+        await create_activity(
```
martastain: Similarly to assignee activities, split the event to individual attributes, throw away calculated attributes (available since https://github.com/ynput/ayon-backend/pull/995 ) and create an activity for each attribute. with data like:
```
{
   "key": "priority",
   "oldValue": "low",
   "newValue": "high",
}
```

(i'd use generic name "key" rather than "attrib" or "attrib name" since we can then add a simple index that will cover data->'key' that could be useful for more activity types)

martastain (changes requested):

ayon_server/activities/event_hook.py
kalisp: Updated according to the comments

ayon_server/activities/event_hook.py
```diff
@@ -75,6 +79,41 @@ async def handle_status_changed(cls, event: "EventModel"):
...
+        assert event.project is not None, "Project is required for activities"
+        entity = await entity_class.load(event.project, event.summary["entityId"])
+
+        payload = event.payload
+        new_values_dict = payload.get("newValue", {})
+        old_values_dict = payload.get("oldValue", {})
+        calculated_attributes = set(payload.get("calculatedAttributes", []))
+
+        origin_link = f"[{entity.name}]({entity_type}:{entity.id})"
+        for key, new_value in new_values_dict.items():
```
martastain: One tiny issue here: this would not track attribute removal (clean up).
When set to none, it is deleted from attributes so it won's show in newValue. You need to iterate over union of keys from both oldvalue and newvalue.

martastain (dismissed):

ayon_server/activities/event_hook.py
kalisp: Oh, makes sense. Updated.

martastain:
It works nicely, as discussed it skips calculated attributes. my last concern is that those changes propagate to inbox. is that expected, @Innders ? It could potentially be quite noisy.

image

Innders:
> It works nicely, as discussed it skips calculated attributes. my last concern is that those changes propagate to inbox. is that expected, @Innders ? It could potentially be quite noisy.
> 
> image

By default I think it should not as it will causes crazy amount of updates. In the future we will get requests to watch for certain attribute changes, but that's out of scope.

kalisp:
I removed adding watchers for `attrib.change` activity types, that seems to be removing them from Inbox. @martastain

martastain:
> I removed adding watchers for `attrib.change` activity types, that seems to be removing them from Inbox. @martastain

That still notify the existing watchers tho... i'll check what i can do in the inbox query. i guess we should filter it out there, but that's an area requiring hazmat suit. stand by

kalisp:
> > I removed adding watchers for `attrib.change` activity types, that seems to be removing them from Inbox. @martastain
> 
> That still notify the existing watchers tho... i'll check what i can do in the inbox query. i guess we should filter it out there, but that's an area requiring hazmat suit. stand by

I don't think so, no `activity_reference` no message in inbox. At least it looked like that when I was testing it. I have a watcher account on folder, only this stopped item showing up in Inbox.

martastain (approved):
@kalisp You are right. Just tested that and it works just fine. Approving

PR #994, opened 2026-06-23
Get latest version with reviewable for multiple entities

### LINKED ISSUE #978: Get latest version with reviewable for entities
It would be quite useful to have an endpoint that returns the latest versions with a reviewable for a set of entities, folders or tasks.

This is basically a clone of the `review-create-session-from-tasks` action from the review addon.

An added bonus would be a field to decide which version to extract from the folder, latest or latest done. A little bit like featured versions.

It would have been very helpful here: https://github.com/ynput/ayon-frontend/pull/2057

kalisp:
## Description of changes

Adds options to query for multiple `folders|products|tasks|versions` at same time.

Adds also new `lastest_done` argument to use only approved versions.

Adds `latest` argument to return latest (even not approved yet) versions.

Using `latest` or `latest_done` arguments will result in returning only versions WITH reviewables. Without it will return even without any reviewables (original implementation).

If multiple versions are returned (for `folder`), it sorts physically latest publishes version first even if there is higher version of different product. It keeps explicit `reviewableOrder` inside of that version (as artists might preferred some reviewable).

### Technical details

It uses `folder_access_list` to return only items that artist has access to. It raises `ForbiddenException` if artist doesn't have access to project at all.

It could be tested by smth like POST to 'http://localhost:5000/api/projects/ayon_project/versions/reviewables/list?latest_done=1' WITH body containing:
```
{
  "entity_ids": [
    "c5d3fcc170a111f1bf59b8f77533e6e9",
  ]
}
```


martastain:
how many entities this is supposed to query at the same time?

kalisp:
> how many entities this is supposed to query at the same time?

Thats probably question for @Innders.

Is the concern size of returned data (so there should be some pagination introduced) (or performance of  searching for list of ids?)

martastain:
Not really. Result is okay, but i can omagine situations where list of IDs won't fit to a query string. Also it loads every entity individually to ensure ACL, which will be very slow in the case of larger lists.

kalisp:
I could change GET to POSTs and put ids to Body. (Or pass some delimiter connected list and parse it out myself, to skip repeating of `*_id`.)

I tried to use asyncio gather to make ACL checks more palatable, not sure if there would be any faster approach.

(But I assume tens or higher tens of ids from where the request came from.)

martastain:
This way, if you don't have access to any of the entities, you get 403. This might be correct (as long you can figure out which entity caused the issue), but if you added a join to hierarchy in the master query and use the folder_access_list there, you could get reviewables for all entities from the list you have access to, without calling ensure_enitity_access first. you'd be down one query and can return a partial response. 

if the intention for this is to be used in review addon, it is possible that a user does not have access to the entire session. in that case, it should probably resolve everything they have access to instead of failing. 

but i like the expansion of ensure_entity_access - if that does not break anything, we should totally keep it

kalisp:
I updated main query to use `folder_access_list` and capture full `ForbiddenException`. Not sure if there should be an option to actually raise it, if nothing is accessible.

Innders:
@kalisp Would mind updating the title and description to be clearer on the goal of the PR. I only know this is related to reviewables because of the attached issue.

api/review/listing.py
```diff
@@ -320,4 +408,32 @@ async def get_reviewables_for_folder(
         project_name,
         folder_id=folder_id,
         user=user,
+        latest_done=latest_done,
     )
+
+
+@router.post("/{entity_type}/reviewables", dependencies=[AllowGuests])
```
martastain: Even tho it does not conflict with anything, this route feels like "create a reviewable for an entity type".

Maybe it should be something like `/{entity_type}/list-reviewables` or `/{entity_type}/reviewables/list`

martastain (changes requested):

api/review/listing.py
kalisp: Changed to `/{entity_type}/reviewables/list` which seems simpler.

martastain (approved):

Innders:
@kalisp What is the best way to get the latest reviewable for the entity. I can see it has latest done, is that the same thing? I just want the latest regardless of the status.

kalisp:
> @kalisp What is the best way to get the latest reviewable for the entity. I can see it has latest done, is that the same thing? I just want the latest regardless of the status.

Added only `latest` argument.

Innders (approved):
Thanks 👍

Innders (changes requested):
Sorry I just found an issue.

When requesting the latest reviewable versions from folders, 1 folder is returning multiple versions. It should always be 1 entity = 1 version. It looks like it is returning multiple task's versions inside the folder. There are 3 tasks in the folder with reviewables.

payload
```json
{
    "entityIds": [
        "eef61391743f11ee9bf5c1fe82e0043e"
    ]
}
```

response
```json
[
    {
        "id": "404fb942414111f095bbc66ef1b47a25",
        "name": "v002",
        "version": "2",
        "status": "data",
        "productId": "5d65e5d4414011f094e4c66ef1b47a25",
        "productName": "audioMain",
        "productType": "audio",
        "attrib": {
            "fps": 24,
            "source": "{root[work]}/wing_it/editorial/editorial.edit.v004.hrox",
            "comment": "",
            "machine": "milan-macbook.local",
            "families": [
                "audio"
            ],
            "handleEnd": 10,
            "handleStart": 10
        },
        "reviewables": []
    },
    {
        "id": "3fa8d636414111f095bbc66ef1b47a25",
        "name": "v003",
        "version": "3",
        "status": "Approved",
        "productId": "3656664c413d11f085b9c66ef1b47a25",
        "productName": "plateMain",
        "productType": "plate",
        "attrib": {
            "fps": 24,
            "source": "{root[work]}/wing_it/editorial/editorial.edit.v004.hrox",
            "comment": "",
            "machine": "milan-macbook.local",
            "families": [
                "plate",
                "clip",
                "review",
                "ftrack",
                "trim"
            ],
            "frameEnd": 1037,
            "handleEnd": 10,
            "colorSpace": "Gamma2.2",
            "frameStart": 1001,
            "handleStart": 10
        },
        "reviewables": []
    },
    {
        "id": "695d84ce424811f08b6dc66ef1b47a23",
        "name": "v004",
        "version": "4",
        "status": "Reviewed",
        "productId": "2163208041f711f0888cc66ef1b47a23",
        "productName": "renderLightingMain",
        "productType": "render",
        "attrib": {
            "fps": 24,
            "source": "/Users/milan/projects/wing_it/resources/shots/comp/010_0020.mov",
            "comment": "",
            "machine": "milan-macbook.local",
            "families": [
                "render",
                "review"
            ],
            "frameEnd": 1037,
            "handleEnd": 0,
            "frameStart": 1001,
            "handleStart": 0
        },
        "reviewables": [
            {
                "fileId": "6cf4f90a424811f0a4f10242ac140002",
                "activityId": "6d1c8664424811f0a4f10242ac140002",
                "filename": "wi_010_0020_renderLightingMain_v004_h264.mp4",
                "label": "h264",
                "mimetype": "video/mp4",
                "availability": "ready",
                "mediaInfo": {
                    "codec": "h264",
                    "width": 1920,
                    "height": 1080,
                    "duration": 1.541667,
                    "frameRate": 24,
                    "iframeOnly": true,
                    "majorBrand": "isom",
                    "pixelFormat": "yuv420p",
                    "probeVersion": 1,
                    "videoTrackIndex": 0
                },
                "createdFrom": null,
                "processing": null,
                "createdAt": "2025-06-05T21:05:28.406309+01:00",
                "updatedAt": "2025-06-05T21:05:28.406309+01:00",
                "author": {
                    "name": "milan",
                    "fullName": "Dean Miller"
                }
            }
        ]
    }
]
```

kalisp:
> When requesting the latest reviewable versions from folders, 1 folder is returning multiple versions.

But that is how the data model looks like. 

Versions belong to the product and there could be multiple products per folder_id. I don't think there is something as 'folder' version.

Innders:
> > When requesting the latest reviewable versions from folders, 1 folder is returning multiple versions.
> 
> But that is how the data model looks like.
> 
> Versions belong to the product and there could be multiple products per folder_id. I don't think there is something as 'folder' version.

There is no such thing as a "Folder" version but the same could said about tasks as well.

The goal the client is to find the latest reviewable version published within the entity. What was the last version I can watch within this folder; product doesn't matter.

The idea is that you can select multiple shots (folders) and create a review session from it.

If the versions are sorted by latest first, I can take the first one and disregard the rest.

kalisp:
I added DB based filter on versions only with `reviewables` which is triggered by using `latest` or `latest_done`.

Without these arguments even versions without reviewables will be returned.

kalisp:
@jakubfiala FYI we decided changing default order of `get_reviewables` methos (`/versions/{version_id}/reviewables`).

jakubfiala:
@kalisp alright, thanks for the heads up! Does this mean we're happy to show them in that order in the UI? Or do I need to change anything in Review?

kalisp:
> @kalisp alright, thanks for the heads up! Does this mean we're happy to show them in that order in the UI? Or do I need to change anything in Review?

I punted the question to @Innders . But as I understood discussion it should be fine.

Innders:
@jakubfiala which order in the UI are we talking about here?

jakubfiala:
@Innders I'm guessing anywhere a list of reviewables is shown, right?

PR #996, opened 2026-06-26
Data Import: copy attribute with multi select

kalisp:
## Description of changes

Without this PR Attributes set as Multi-Select couldn't be imported.

### Technical details

The error would be: (as multi-select type was resolved as 'string' and not `list_of_strings`)
```
attrib -> reel
  value is not a valid list (type=type_error.list)
 ```
 
 I tested it on smth like:
 
image


martastain:
That error you are mentioning is correct. In the instance this attribute is used, it is misconfigured and exects a list. so the import should fail for simple strings. i am missing something?

kalisp:
Ordinary multi-select attributes couldnt be imported before as it was pulling wrong data type (`str` instead of `list_of_strings`)

PR #1001, opened 2026-06-30
Data Import: Remove unnecessary events, keep updating first one

kalisp:
## Description of changes

Data imports create too many progress events which are causing slowdown and possible service degradation. 

This should set up single event and keep updating it.

### Technical details

Current look of `summary` field is necessary for FE as it is depending on calculation of progress.

Should be tested together with:
|https://github.com/ynput/ayon-frontend/pull/2111
and also could with:
https://github.com/ynput/ayon-backend/pull/996


kalisp:
Still WIP:
Issues:
- frontend disappears that event in Event Viewer after first update with store=False, it reappears after is finished
- data import FE expects event on import.data.finish after full phase, which would require keeping existing finishing event there (and have last update of starting event next to it which was butt ugly)
- the most important problem - it doesn't help FE with Refresh issue, it still show up that notification

martastain:
Could you pls add  logger.debug to the end, with information on how long the processing took and ideally also average duration of an operation? :pray:  i think that would be really useful for comparison you know where.

kalisp:
> Could you pls add logger.debug to the end, with information on how long the processing took and ideally also average duration of an operation? 🙏 i think that would be really useful for comparison you know where.

image

martastain (approved):

PR #1004, opened 2026-07-02
Query throttle

martastain:
This pull request introduces a concurrency throttling mechanism to limit the number of simultaneous requests a user can make for specific API endpoints, helping to prevent abuse and improve system stability. The throttling is enforced per user session using Redis, and a new exception is added to handle cases where the limit is exceeded. The most important changes are grouped below.

* Added a `user_request_throttler` async context manager in `ayon_server/api/auth.py` that enforces per-user concurrent request limits for GraphQL and operations endpoints, using Redis to track counts. If the limit is exceeded, a `TooManyRequestsException` is raised.
* Integrated the request throttler into the `AuthMiddleware` so that every request checks concurrency limits before processing.
* Introduced a new `TooManyRequestsException` in `ayon_server/exceptions.py` to represent HTTP 429 errors when users exceed allowed concurrent requests.
* Added a `decr` method to the `Redis` class in `ayon_server/lib/redis.py` to decrement counters, supporting the throttling logic.


kalisp:
Could we mention here what are the limits, for future posterity? Or would they be configurable?
(I see 10 same graphQL queries per user and 2 (sets??) of operations?)

(And maybe how this could get tested? I tried to run 3 different Data Imports at same time, which should produce 3 sets of operations, didn't see `TooManyRequestsException` in logs.)

BigRoy:
This may require desktop tools to respond to this in the right reasonable manner as well. We should probably throttle or delay the retries when we face this response. @iLLiCiTiT

martastain:
>  I tried to run 3 different Data Imports at same time, which should produce 3 sets of operations, didn't see TooManyRequestsException in logs.)

Import is now unaffected, it would make sense to add another rule with limit 1. It just uses operations internally, so this middleware won't catch it

> (And maybe how this could get tested? I tried to run 3 different Data Imports at same time, which should produce 3 sets of operations, didn't see TooManyRequestsException in logs.)

ATM this is an experiment - before merging, i'd raise limits to sane levels. technically you should never reach the state when a single session triggers more than 10 graphql requests.... same goes for operations. If you do so, it's probably because of a bug. Especially in the case of the frontend - i could actually group by sender if that makes sense.

PR #1010, opened 2026-07-07
Fix list guest access

martastain:


Innders (changes requested):
`activityCategories` is always an empty list for guests

Innders (approved):

PR #1011, opened 2026-07-07
Modified stats for inherited attribs to collect from all levels

kalisp:
## Description of changes

This PR modifies calculation of stats for columns containing `inherited_attibutes` to collect values from `item.attrib`, `item.inherited_attributes` and `project_attributes` (if available). (In this order of precedence.)

This should handle wrong numbers for fields like `priority`.

### Technical details

image (this PR)

vs

image (previous)

It is assumed that if we are asking for 'inherited' values, we want to actually get all levels. If there would be a need to differentiate between each level, new name (`effective.priority`) would need to be introduced in frontend (and BE updated accordingly).

(Now FE sends `inherited.priority` even if it means all levels (for example).

### DIFF
--- a/ayon_server/graphql/resolvers/field_stats.py
+++ b/ayon_server/graphql/resolvers/field_stats.py
@@ -112,7 +112,10 @@ def _get_stats_for_column(
     return []
 
 
-def generate_specific_stats_columns(calculate_specific_statistics) -> str:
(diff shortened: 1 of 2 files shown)


Innders (approved):
Seems to work for me 🥳

martastain (approved):

PR #1012, opened 2026-07-09
feat: better logging in auto-update

martastain:


martastain:
Addressed in a different pr

PR #1015, opened 2026-07-10
Fix classmethod request coalescing

martastain:
This pull request improves the `RequestCoalescer` utility to better handle coalescing of asynchronous requests, especially for class methods and instance methods, and refactors project entity loading to leverage these improvements. It also adds comprehensive tests to ensure correct batching and coalescing behavior.

### Request coalescing improvements

* Enhanced the `_hash_args` function in `request_coalescer.py` to generate stable, unique identifiers for bound methods (class and instance methods), ensuring that coalescing works correctly for these cases.
* Refactored the batching logic in `RequestCoalescer.__call__` to allow multiple batches when the number of waiters exceeds `max_waiters`, ensuring that excess requests are grouped into new batches with unique keys.
* Updated `ProjectEntity.load` in `project.py` to use the improved `RequestCoalescer` for coalescing database loads, and split the actual load logic into a new `_load` method for clarity and flexibility.
* Removed direct usage of `RequestCoalescer` in `api/projects/projects.py` in favor of the new coalescing logic built into `ProjectEntity.load`.

### Testing

* Added tests to `test_ayon_server_utils.py` that verify coalescing works for class methods, instance methods, and that overflow waiters are properly batched into separate tasks when exceeding `max_waiters`.

### DIFF
--- a/ayon_server/entities/project.py
+++ b/ayon_server/entities/project.py
@@ -26,8 +26,7 @@ from ayon_server.helpers.inherited_attributes import rebuild_inherited_attribute
 from ayon_server.helpers.project_list import build_project_list
 from ayon_server.lib.postgres import Postgres
 from ayon_server.lib.redis import Redis
-from ayon_server.utils import SQLTool, dict_exclude, get_nickname
-from ayon_server.utils.request_coalescer import RequestCoalescer
+from ayon_server.utils import RequestCoalescer, SQLTool, dict_exclude, get_nickname
 
 if TYPE_CHECKING:
     from .project_skeleton import ProjectSkeletonEntity
--- a/ayon_server/entities/project.py
+++ b/ayon_server/entities/project.py
@@ -27,6 +27,7 @@ from ayon_server.helpers.project_list import build_project_list
 from ayon_server.lib.postgres import Postgres
 from ayon_server.lib.redis import Redis
 from ayon_server.utils import SQLTool, dict_exclude, get_nickname
+from ayon_server.utils.request_coalescer import RequestCoalescer
 
 if TYPE_CHECKING:
     from .project_skeleton import ProjectSkeletonEntity
@@ -67,7 +68,20 @@ class ProjectEntity(TopLevelEntity):
     async def load(
         cls,
         name: str,
-        transaction=None,  # deprecated
+        transaction: Any = None,  # deprecated
+        for_update: bool = False,
+    ) -> "ProjectEntity":
+        """Load a project from the database."""
+
+        if not for_update:
+            coalesce = RequestCoalescer()
+            return await coalesce(cls._load, name)
+        return await cls._load(name, for_update=for_update)
+
+    @classmethod
+    async def _load(
+        cls,
+        name: str,
         for_update: bool = False,
     ) -> "ProjectEntity":
         """Load a project from the database."""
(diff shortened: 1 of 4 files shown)


ayon_server/utils/request_coalescer.py
```diff
@@ -11,7 +11,21 @@ def _hash_args(func: Callable[..., Any], *args: Any, **kwargs: Any) -> str:
...
+            func.__self__
+            if isinstance(func.__self__, type)
+            else func.__self__.__class__
+        )
+        func_name = f"{self_class.__module__}.{self_class.__qualname__}.{func.__func__.__name__}"  # noqa: E501
+        func_id = f"{func_name} (bound to {id(func.__self__)})"
+    elif hasattr(func, "__qualname__"):
+        func_id = f"{getattr(func, '__module__', '')}.{func.__qualname__}"
+    else:
+        func_id = f"{str(func)} (id: {id(func)})"
```
BigRoy, line 27: Oh boy :') Feels like something `inspect` library would be able to do, but I guess not?

ayon_server/entities/project.py
```diff
@@ -67,7 +68,20 @@ def return_project_skeleton(
     async def load(
         cls,
         name: str,
-        transaction=None,  # deprecated
+        transaction: Any = None,  # deprecated
+        for_update: bool = False,
```
BigRoy, line 71: Should we enforce `for_update` to be a keyword only argument?
The only reason is to keep space for removing deprecated `transaction` argument at some point. Because it'd enforce that no one would pass three positional arguments. Or is that overthinking things?
```suggestion
        *,
        for_update: bool = False,
```

BigRoy (commented):
I have no idea what this code is doing so it'll be hard to provide any reasonable comments. 🗡️ 

Do you need me to test whether the breaking query now passes for me too?

ayon_server/entities/project.py
martastain: Yeah. i am just scared it is used in some addon in the wild as positional. the same goes for "transaction" that is deprecated for a year

iLLiCiTiT: Well, `for_update` is not used as positional argument anywhere, or?

martastain: I hope not. But.... if i was to break it, i'd break it at least with the next minor bump (soon)

ayon_server/utils/request_coalescer.py
martastain: not really. i tried, then gemini tried..... and this works, even if it's ugly AF (clanker's code, but kinda makes sense. apart from ugliness)

ayon_server/entities/project.py
iLLiCiTiT: This PR adds `for_update`, it can't break anything. Or what do I miss?

martastain: for_update was always there. it's just the diff that dows not show it. 

old `load` was changed to `_load` and new `load` just wraps it and handles request coalescing. same signature

iLLiCiTiT (approved):

ayon_server/entities/project.py
BigRoy: Ah you're right - it's the diff that made it look confusing. Ignore then.

PR #1016, opened 2026-07-10
Bugfix: enhance task filtering to include inherited attributes

### LINKED ISSUE #1014: Stats: Folder stats with inherited values not taken into account
## Story

I'm filtering on some tasks and search by an inheritable field like "priority". I haven't explicitly set some tasks priority value so they are inherited from the parent folder.

### To reproduce

1. Find an inherited attrib like `priority`
2. Filter on the **task** for that attrib like `priority: high`
3. The folder stats count will be wrong as it does not count the inherited values

### Expected behavior

The `taskFilter` on folders `search` will return the folders that still have tasks matching the filter in them (not empty). The filter will match inherited task attrib values and include those folders.

```json
{
    "taskFilter": {
        "conditions": [
            {
                "key": "attrib.priority",
                "value": [
                    "high"
                ],
                "operator": "in"
            }
        ],
        "operator": "and"
    },
    "search": ""
}
```

The folder stats resolver should work in the same way and take into account inherited task values.

### Additional context

Image

kalisp:
## Description of changes

This handles weird issue in `Folder` counts in Summary row if task filtering is used (on `priority` = 'high') for example.

Original query builder looked for folders which had tasks with `priority` = 'high', but if 'high' was inherited from `folder`, that `folder` wasn't included and it wasn't calculated in `Summary` row.

### Technical details
Original:

Screenshot 2026-07-10 131048

(See `Folders: 0` in left bottom)

This PR:

image

### Additional context

This needs to be tested thoroughly as it is touching query builder!

### DIFF
--- a/ayon_server/graphql/resolvers/folders.py
+++ b/ayon_server/graphql/resolvers/folders.py
@@ -579,22 +579,24 @@ async def get_folders(
 
         fdate = json.loads(task_filter)
         fq = QueryFilter(**fdate)
-        # Build a single filter that coalesces the attribute value across
-        # the task itself, its inherited (exported) attributes and the
-        # project attributes. A folder matches if any of its tasks has the
-        # requested attribute value in any of those sources.
-        task_filter_cond = build_filter(
+        tfilter = build_filter(
             fq,
             column_whitelist=column_whitelist,
             table_prefix="tasks",
-            column_map={
-                "attrib": (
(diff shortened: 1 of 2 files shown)


Innders (approved):
Seems to be fixed for me

kalisp:
Removed updates in sql builder as they were unnecessary (and possibly dangerous). Updated task filter query with use of  `column_map` should be enough.

@Innders could you please retest it?

Innders (approved):

martastain (approved):

PR #1017, opened 2026-07-10
Fix chunked upload to S3

martastain:
- Amazon S3 allows uploading parts of varying sizes as long as all parts (except the last one) are at least 5 MB.
- Cloudflare R2 strictly requires all non-trailing parts of a multipart upload to be exactly the same size. 
- In s3.py, incoming chunks from the upload request stream and remote downloads were buffered and pushed as soon as  `len(buff) >= buffer_size`  (5 MB). Because stream chunk sizes vary, this created non-last parts of slightly different sizes (e.g.  5 MB + X bytes ), triggering the  InvalidPart  error.

##  The Fix
- Updated the buffering logic in s3.py and s3.py in s3.py to slice the buffer.
- Every time  len(buff) >= buffer_size , the exact first  buffer_size  bytes are pushed as a part, and the remainder is carried over to the next loop iteration. This guarantees that all non-last parts are exactly  buffer_size  in length, leaving only the very last chunk to be smaller.

> [!WARNING]  
> This must be tested on a cloud instance before merging!


64qam (approved):
LGTM

PR #1019, opened 2026-07-14
Hide support users

martastain:
This pull request updates the user visibility logic to improve support user privacy and deprecates the `is_support` filter in the users GraphQL resolver. The main focus is to ensure that support users are only visible to other support users, and to simplify the filtering logic accordingly.


BigRoy (approved):
Logic makes sense - not sure how I'd test this with my dev server though.

Innders (commented):
Not testable and I don't know ~~code~~ python so I'll remove myself here

martastain:
Tested in cloud, works

PR #1021, opened 2026-07-14
Feature: fuzzy search on entity_list_items

### LINKED ISSUE #1008: Lists: Search for Entity List Items
## Story

There is a particular shot I want to find in my review session.
I only want to see tasks being with the word "light".

### Problems

You cannot use a fuzzy search on entity list items and it is inconsistent with the rest of the entity types.

### Proposal

Add a `search` argument on `entityList.items` graphql that will filter on the `name`, `label`, `path` of the entity list item.

kalisp:
## Description of changes

This PR adds `search` argument for fuzzy string search on `entity_list_items` graphQL query.

It searches for string values in:
- item labels
- item names
- item paths

(Not sure if this search is enough, open for expanding it after comments.)


ayon_server/graphql/resolvers/entity_list_items.py
```diff
@@ -571,6 +572,36 @@ async def get_entity_list_items(
...
+            hierarchy_path_col = None
+
+        if entity_type == "version":
+            sql_columns.append("pd.name AS _search_entity_name")
+        elif entity_type == "workfile":
+            sql_columns.append("e.path AS _search_entity_name")
+        else:
+            sql_columns.append("e.name AS _search_entity_name")
+
+        terms = slugify(search, make_set=True, min_length=2)
```
filipvnencak: Do we want to silently ignore 1-character searches?

ayon_server/graphql/resolvers/entity_list_items.py
```diff
@@ -571,6 +572,36 @@ async def get_entity_list_items(
...
+        for term in terms:
```
filipvnencak: Terms are ANDed here but Overview search (tasksFolders resolver) splits on comma so chips are OR'd. Same search UI, different behavior.

ayon_server/graphql/resolvers/entity_list_items.py
filipvnencak: slugify runs unidecode on the query, but names/labels can be saved in a different script (Cyrillic, CJK...), those will never match. Just pointing out, in case we're aware of it.

filipvnencak (changes requested):

ayon_server/graphql/resolvers/entity_list_items.py
kalisp: Standardized to `folders` behavior.

ayon_server/graphql/resolvers/entity_list_items.py
kalisp: Removed limit.

kalisp: I dont think this is a problem. Internally it is stored in utf, imo.

martastain: But slugify transliterates non-ascii characters to ascii.... it is a good point actually. but i would say it is a temporary problem as we aim to use native fulltext soon and this hack will go away

kalisp: Daamn, you are right. I tested it wrong.

image

martastain: In the ideal world, text vector would be extracted from text fields, transliterated to ascii and stored separatelly. Then, when you transliterate the search term, it enables you to have character-set-agonstic search. but we don't have that index yet.

filipvnencak (changes requested):
Way better! But I noticed the WHERE conditions are missing parentheses. Since AND has higher precedence than OR, the search conditions after the first OR escape the entity_list_id scoping, I got items leaking in from other lists, and multi-word search behaves like OR instead of AND.

Note: folders.py, tasks.py and entity_lists.py have the same issue at the comma level, their t1_conds OR-join is also appended without parens.

ayon_server/graphql/resolvers/entity_list_items.py
```diff
@@ -571,6 +572,47 @@ async def get_entity_list_items(
...
+                    f"label ILIKE '%{term}%'",
+                    f"_search_entity_name ILIKE '%{term}%'",
+                ]
+                if hierarchy_path_col is not None:
+                    sub_conditions.append(f"{hierarchy_path_col} ILIKE '%{term}%'")
+
+                t2_conds.append(
+                    SQLTool.conditions(sub_conditions, "OR", add_where=False)
+                )
+            t1_conds.append(SQLTool.conditions(t2_conds, "AND", add_where=False))
```
filipvnencak: `SQLTool.conditions` doesn't add parentheses, so this ends up as `A OR B OR C AND D OR E OR F`, AND is evaluated first and the groups fall apart. `modeling characters` gives 11 rows instead of 4, and `ep101 ep104` vs `ep104 ep101` return different rows.

```suggestion
                t2_conds.append(
                    f"({SQLTool.conditions(sub_conditions, 'OR', add_where=False)})"
                )
            if t2_conds:
                t1_conds.append(
                    f"({SQLTool.conditions(t2_conds, 'AND', add_where=False)})"
                )
```

ayon_server/graphql/resolvers/entity_list_items.py
```diff
@@ -571,6 +572,47 @@ async def get_entity_list_items(
...
+
+        sql_conditions.append(
+            f"({SQLTool.conditions(t1_conds, 'OR', add_where=False)})"
+        )
```
filipvnencak: A search of only special characters (`?`) loses every term in slugify and this appends `()` → `AND ()` → 500. In the UI it shows up as "corrupt filter".

```suggestion
        if t1_conds:
            sql_conditions.append(
                f"({SQLTool.conditions(t1_conds, 'OR', add_where=False)})"
            )
```

ayon_server/graphql/resolvers/entity_list_items.py
```diff
@@ -571,6 +572,47 @@ async def get_entity_list_items(
...
+        else:
+            sql_columns.append("e.name AS _search_entity_name")
+
+        parts = search.split(",")
+        t1_conds = []
+        for part in parts:
+            terms = slugify(part, make_set=True, split_chars=" ")
+            t2_conds = []
+            for term in terms:
+                term = term.replace("'", "''")  # Escape single quotes
```
filipvnencak: slugify strips `'` before this, so it never fires. `_` and `\` do survive though - `sh_010` matches `shX010` too.

filipvnencak (changes requested):

ayon_server/graphql/resolvers/entity_list_items.py
kalisp: applied

ayon_server/graphql/resolvers/entity_list_items.py
```diff
@@ -571,6 +572,49 @@ async def get_entity_list_items(
...
+                t2_conds.append(
+                    f"({SQLTool.conditions(sub_conditions, 'OR', add_where=False)})"
+                )
+            if t2_conds:
+                t1_conds.append(
+                    f"({SQLTool.conditions(t2_conds, 'AND', add_where=False)})"
+                )
+
+        if t1_conds:
+            sql_conditions.append(SQLTool.conditions(t1_conds, "OR", add_where=False))
```
filipvnencak: Cross lists leaking enitities 


Only one Task list search (4 item): 
Image

New List with same tasks added and search returning 2 more Tasks: 
Image

```suggestion
            sql_conditions.append(
                f"({SQLTool.conditions(t1_conds, 'OR', add_where=False)})"
            )
```

filipvnencak (changes requested):

filipvnencak (approved):
Works, thanks!

Innders:
Great, @kalisp please merge!

PR #1031, opened 2026-07-28
YnputCloud disconnect

martastain:
This pull request updates the process for removing the Ynput Cloud key to ensure that the instance is properly disconnected from the Ynput Cloud when an admin deletes the key. The main changes include updating the API endpoint to trigger a disconnect, modifying the helper method to support this behavior, and improving error handling and logging.

*  The `delete_ynput_cloud_key` endpoint now disconnects the instance from the Ynput Cloud in addition to removing the key. The admin-only restriction message has been clarified.
* `remove_ynput_cloud_key` method now accepts a `disconnect_instance` parameter. If set, it attempts to notify the Ynput Cloud API to disconnect the instance, with error handling and logging if the disconnect fails.

### DIFF
--- a/api/ynputcloud/connect.py
+++ b/api/ynputcloud/connect.py
@@ -81,8 +81,11 @@ async def set_ynput_cloud_key(
 
 @router.delete("")
 async def delete_ynput_cloud_key(user: CurrentUser) -> EmptyResponse:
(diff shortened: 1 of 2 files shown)


martastain:
> Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.

I didn't request it you sucker!

PR #1037, opened 2026-07-30
Profile endpoints

### LINKED ISSUE #1034: Login Sessions: allow changing the `fullName` in a guest login session
## Story

As a guest using a public link to e.g. a review session, I want to set my display name so that others can identify my comments, annotations and sync cursors (among other things). To do this, my login session must be modified in the backend, replacing the default `Anonymous Guest` with my chosen name.

### Problems

This will allow us to add name input fields for guest users.

### Proposal

add an endpoint to modify the attributes of the current login session in Redis. Probably just allow changing `fullName` for now.

## TODO

- [ ] ...

### LINKED ISSUE #1036: Activities: Store the author's `fullName` in the activity for guests
## Story

As a guest user, I want other users to identify my comments and other activities using my chosen name.

### Problems

Currently the full names shown in the activity feed are pulled from the database. For anonymous guests (i.e. those whose full name isn't stored in project data), this means we cannot retrieve any full name for them and we default to "Anonymous Guest".

### Proposal

When submitting/patching an activity **created by an anonymous guest**, we should grab the `fullName` attribute from their login session and store it in the activity data structure itself.

When returning the activity in the GraphQL endpoint, we should set the `authorFullName` field to this value, if (and only if) available.

IMPORTANT: this should only happen for guests for whom we don't already have a full name available in project data. Otherwise the field on the activity should be empty and we should retrieve the full name from the users table / project data.

## TODO

- [ ] ...

martastain:
This pull request introduces a new `/api/profile` endpoint for user profile management, deprecates the old `/api/users/me` endpoint, and adds improved handling for guest and project guest users. The changes are organized around API improvements, guest user handling, and internal refactoring.

The new "update_current_user_profile" additionally supports updating anonymous guest users session data, so they can update their own full names in order to be identified later.

image

image

### DIFF
--- a/api/profile/profile.py
+++ b/api/profile/profile.py
@@ -1,0 +1,93 @@
+from typing import Any
+
+from ayon_server.api.dependencies import (
+    AllowGuests,
+    CurrentUser,
+)
+from ayon_server.auth.session import Session
+from ayon_server.entities import UserEntity
+from ayon_server.exceptions import (
+    BadRequestException,
+)
+from ayon_server.lib.redis import Redis
+from ayon_server.logging import logger
+
+from .router import router
+
+#
+# [GET] /api/users/me
+#
+
+
+@router.get("", dependencies=[AllowGuests])
+async def get_current_user_profile(
+    user: CurrentUser,
+) -> UserEntity.model.main_model:  # type: ignore
+    """
+    Return the current user information (based on the Authorization header).
+    This is used for a profile page as well as as an initial check to ensure
+    the user is still logged in.
+    """
+
+    payload = user.payload
+    payload.ui_exposure_level = await user.get_ui_exposure_level()  # type: ignore
+    payload.data.pop("supportToken", None)  # type: ignore
+    return payload
+
+
+async def update_guest_attrib(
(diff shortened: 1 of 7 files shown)


api/profile/profile.py
```diff
@@ -0,0 +1,93 @@
...
+from ayon_server.exceptions import (
+    BadRequestException,
+)
+from ayon_server.lib.redis import Redis
+from ayon_server.logging import logger
+
+from .router import router
+
+#
+# [GET] /api/users/me
```
jakubfiala, line 18: This should be `[GET] /api/profile`, right?

jakubfiala (approved):
LGTM. Once this is merged I'll open an accompanying PR in ayon-frontend to use the `GET /api/profile` endpoint instead of `/api/users/me`. Just got one clarifying question.

api/profile/profile.py
martastain: sure. i'll fix that. it's a result of copy-paste.

This is a groundwork - the patch endpoint will need more love, this is just a MVP to solve the problem with guests

PR #1040, opened 2026-07-30
Import Data: Finish import and not wait for events

kalisp:
## Description of changes

Full processing of events could be expensive, for importer not so important. It finishes importing of data as soon as data are in DB, not after all events triggered by updates are.


BigRoy (approved):
Didn't test but LGTM

PR #1041, opened 2026-07-31
Import data: Validation phase speedup

kalisp:
## Description of changes

Implemented various steps to mitigate slowness of validation phase.

### Technical details

Now it is so fast (on my machine) that FE cannot keep up with redrawing progress bar :).

### DIFF
--- a/api/data_import/import_data.py
+++ b/api/data_import/import_data.py
@@ -995,7 +995,7 @@ async def _resolve_entity_id(
                 path_to_ids[path] = entity_id  # Cache it
                 return entity_id
         except NotFoundException:
-            logger.debug(f"Couldn't find entity for path '{path}'")
+            logger.trace(f"Couldn't find entity for path '{path}'")
 
     return None
 
--- a/api/data_import/import_data.py
+++ b/api/data_import/import_data.py
@@ -263,13 +263,7 @@ async def import_data(
                 project=project_name,
                 description=f"Validating row: {row_number} of {total_rows} rows",
                 progress=current_progress,
-                summary={
-                    "created": import_status.created,
-                    "updated": import_status.updated,
-                    "skipped": import_status.skipped,
-                    "failed": import_status.failed,
-                    "phase": import_status.phase,
-                },
+                summary=await _prepare_status_summary(import_status),
                 status="in_progress",
                 store=False,
             )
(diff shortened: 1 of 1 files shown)


BigRoy:
@martastain can we test this on your secret big database clone?

api/data_import/import_data.py
BigRoy: If this is now indeed lightning fast somehow - should we skip the emit if it's not the last percentage and the last emit was e.g. 0.01s ago? Just to avoid emitting 100 times in say 0.1s? (Emitting less events might even speed it up more?)

kalisp: We are just updating single event, not emitting.

BigRoy: Isn't it still the case frontend gets each update through websockets. Aren't we still emitting 100 of these to all connected clients (or maybe only to the one?)
If just the one, maybe not really an issue.

kalisp: We might be pushing 100 update to single user, I had it per 5% before, but @martastain said 100 is fine.

BigRoy (commented):
Giving this PR a go it seems that my frontend just remains on 0% for a long time, about 10 seconds? and then just updates to be done. The progress bar does not move at all but it's not instant either.

So during this green bar I see nothing for the 15 seconds - and then suddenly it completes in the dialog.

Image

@kalisp any idea?

BigRoy:
So the progress bar doesn't move for me because ALL rows for me a "Skipped" because I don't have matching entities on my end - the behavior is the same on the latest server releases though.

It's because it reaches the `except` for me, but it doesn't stop the response and also doesn't emit any progress if it doesn't enter this if statement: https://github.com/ynput/ayon-backend/blob/65b2d61adec5c5aefeca74b755de78d8129d83cb/api/data_import/import_data.py#L409-L430

BigRoy:
Also, the majority of the time taken on this scenario is the logging to the server.


With the log: 14.7s
Without the log: 3s
(It's likely the traceback formatting that's so slow)

So if we can avoid this log it'd already speed things up: https://github.com/ynput/ayon-backend/blob/65b2d61adec5c5aefeca74b755de78d8129d83cb/api/data_import/import_data.py#L402

We could e.g. do
```
if log_level == "trace":
    # Then log it

```
And probably only if it's the `ImportRowErrorException` then always log it? (That runs once only anyway.)

kalisp:
Moved progress logging at the beginning to not having to handle different case of failures.
Bumped traceback logging to `trace` (and actually resolving lazily to have any impact).

(Refactored some most ugliest progress parts.)

api/data_import/import_data.py
```diff
@@ -1081,3 +1093,31 @@ def _to_bool(value: Any) -> bool:
...
+        "created": import_status.created,
+        "updated": import_status.updated,
+        "skipped": import_status.skipped,
+        "failed": import_status.failed,
+        "phase": import_status.phase,
+        "failedItems": import_status.failed_items,
+    }
+
+
+def _trigger_status_update(index: int, total: int) -> tuple[int, bool]:
```
BigRoy, line 1111: Could use a docstring explaining what the return values mean.

BigRoy (approved):
LGTM

PR #1047, opened 2026-08-05
Fix sorting by Folder Name in tasks resolver

### LINKED ISSUE #1046: YN-0988: Overview: Sorting by folder name does not work

Sorting tasks by "folder" or "folderName" in graphql does not work.

```gql
    tasks(sortBy: "folder") {}
```

```json
    { "data": null, "errors": [ { "message": "Invalid sort_by value: folder", "locations": [ { "line": 5, "column": 5 } ], "path": [ "project", "tasks" ], "extensions": { "status": 400 } } ] }
 ``` 

NOTE: The frontend uses "folder" key but looking at versions sorting "folderName" is used.

``` gql
    versions(sortBy: "folderName") {}
 ```

Image

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0988]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=29078cc0909a11f19196417d12b59859)

### PR DESCRIPTION
**Pull request by kalisp**
## Description of changes

PR solves issue when particular folder is used in `Slicer` in `Overview` tab and then subfolders (and tasks) are sorted by `Folder Name`.

### Technical details

image

### DIFF
--- a/ayon_server/graphql/resolvers/tasks.py
+++ b/ayon_server/graphql/resolvers/tasks.py
@@ -58,7 +58,7 @@ SORT_OPTIONS = {
     "updatedAt": "tasks.updated_at",
     "createdBy": "tasks.created_by",
     "updatedBy": "tasks.updated_by",
-    "folder": "folders.name",
+    "folderName": "folders.name",
 }
 
 
--- a/ayon_server/graphql/resolvers/tasks.py
+++ b/ayon_server/graphql/resolvers/tasks.py
@@ -512,7 +512,7 @@ async def get_tasks(
     #
 
     # Do we need the parent folder data?
-    if use_folder_query or "folder" in fields or sort_by == "folder":
+    if use_folder_query or "folder" in fields or sort_by == "folderName":
         sql_columns.extend(
             [
                 "folders.id AS _folder_id",
--- a/ayon_server/graphql/resolvers/tasks.py
+++ b/ayon_server/graphql/resolvers/tasks.py
@@ -58,6 +58,7 @@ SORT_OPTIONS = {
     "updatedAt": "tasks.updated_at",
     "createdBy": "tasks.created_by",
     "updatedBy": "tasks.updated_by",
+    "folder": "folders.name",
 }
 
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Innders** [comment]
Does the query need to use `sortBy: "folderName"` ?

**Comment by kalisp** [comment]
It expects (or as it was provided) 'folder'.

**Review by Innders — COMMENTED** [review]
The fix works but should it not follow the naming in these files?

```python
"folderName": "folders.name",
```
[products.py](https://github.com/ynput/ayon-backend/blob/089bf3e56dc010a672630de36e4beba07db1d421/ayon_server/graphql/resolvers/products.py#L46)
[versions.py](https://github.com/ynput/ayon-backend/blob/089bf3e56dc010a672630de36e4beba07db1d421/ayon_server/graphql/resolvers/versions.py#L53)

kalisp:
> The fix works but should it not follow the naming in these files?

I could change that definitely, but wouldn't that require change in FE too? As now it comes from FE as 'folder', not 'folderName'.

Innders:
> I could change that definitely, but wouldn't that require change in FE too? As now it comes from FE as 'folder', not 'folderName'.

Frontend change is no problem, better to do it right in both places 👍

kalisp:
Changed to 'folderName'.

ayon_server/graphql/resolvers/tasks.py
```diff
@@ -511,7 +512,7 @@ async def get_tasks(
     #
 
     # Do we need the parent folder data?
-    if use_folder_query or "folder" in fields:
+    if use_folder_query or "folder" in fields or sort_by == "folder":
```
Innders: It's not working, maybe because of this line?

Innders (changes requested):

ayon_server/graphql/resolvers/tasks.py
kalisp: Yeah, I ffed up commit&push, sorry.

Innders (approved):

PR #1050, opened 2026-08-06
Speedup of versions resolver

kalisp:
## Description of changes

Experimental PR for `versions` resolver speedup.

### Technical details

It seems that CTEs (for latest, hero versions) are run unnecessarily on full `versions` table even if only page of values should be returned. This could cause significant slowdown in really large projects (as `demo_Big_Episodic`).

CTEs were replaced by LATERAL joins which run only on limited set of rows.

This is HIGHLY experimental PR, for discussion mainly. 

It is possible that CTE approach is preferred, it might be possible that this will ~~break `stats` calculation for columns~~ (on first look it didn't - actually it shouldn't as `stats` queries are missing order and limits, so it doesn't really matter if it is CTE or LATERAL).

### DIFF
--- a/ayon_server/graphql/resolvers/versions.py
+++ b/ayon_server/graphql/resolvers/versions.py
@@ -358,70 +358,53 @@ async def get_versions(
     # Always-on CTEs (to get latest and hero versions)
     #
 
-    sql_cte.extend(
-        [
-            f"""
-            latest_versions AS (
-                SELECT DISTINCT ON (product_id) id, version, product_id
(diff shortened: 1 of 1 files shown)


Innders (commented):
Initial testing on big_demo_episodic getting 250 versions sorting by path shows:

Before: ~2.2s
After: ~300-400ms

5x faster!

martastain:
Now a part of https://github.com/ynput/ayon-backend/pull/1042

PR #1051, opened 2026-08-06
Fuzzy Search: Extracted universal function, updated occurrences

### LINKED ISSUE #1032: Search: Standardize search behavior
## Story

Different `graphQL` resolvers handle `Search` content differently.

Some are parsing it first by ',' and then slugify:
- `tasks`
- `entityLists`
- `folders`

Some just slugify:
- `entityListItems`
- `products`,
- `representations`
- `versions`
- `workfiles`

### To reproduce

1. search content with ',' should be behaving differently in different resolvers

### Expected behavior

This should be synchronized to single approach, if it is not by design. In that case it could be somewhere documented.

## Environment

1.16.1+DEVELOP

kalisp:
## Description of changes

This PR tries to standardize fuzzy search (by names, text values etc) and use reusable methods. There were 2 approaches for handling search terms (with splitting by ',' first or not), this standartizes it to split by ',' always.

It removes limitation on search term needing to be longer than 2 characters.


ayon_server/graphql/resolvers/common.py
```diff
@@ -275,3 +276,45 @@ def get_has_links_conds(
...
+    """
+    parts = search.split(",")
+    t1_conds = []
+
+    for part in parts:
+        part = part.replace("'", "''")
+        terms = slugify(part, make_set=True, split_chars=" ", min_length=min_length)
+        t2_conds = []
+        for term in terms:
+            sub_conditions = [f"{col} ILIKE '%{term}%'" for col in columns]
```
filipvnencak, line 304: `_` survives slugify here (with `split_chars=" "` the other separators get whitelisted) and is a LIKE wildcard, so `shot_01` also matches `shot101`.

```suggestion
            term = term.replace("\\", "\\\\").replace("_", "\\_")
            sub_conditions = [f"{col} ILIKE '%{term}%'" for col in columns]
```

ayon_server/graphql/resolvers/common.py
```diff
@@ -275,3 +276,45 @@ def get_has_links_conds(
...
+    Within each term, the specified columns are OR'd together.
+
+    Returns a SQL condition string, or None if search is empty
+    or no conditions could be built.
+    """
+    parts = search.split(",")
+    t1_conds = []
+
+    for part in parts:
+        part = part.replace("'", "''")
```
filipvnencak: slugify drops `'` anyway (not in the whitelist), so this escape never reaches the SQL, the sanitising is really the whitelist's job. Keep it with a note, or drop it?

filipvnencak (changes requested):

ayon_server/graphql/resolvers/common.py
```diff
@@ -275,3 +276,45 @@ def get_has_links_conds(
...
+    Within each term, the specified columns are OR'd together.
+
+    Returns a SQL condition string, or None if search is empty
+    or no conditions could be built.
+    """
+    parts = search.split(",")
+    t1_conds = []
+
+    for part in parts:
+        terms = slugify(part, make_set=True, split_chars=" ", min_length=min_length)
```
martastain: maybe we could just use 

```python
terms = part.lower().replace("'","''").split(" ")
```

 here and have non-ascii matching as @filipvnencak pointed out somewhere. it doesn't do much else with these params, apart from transliteration, which actually may cause issues rather than help.

kalisp: Replaced `slugify` with this.

It didn't help with nonstandard (cyrilic) characters though. (Even if it should as query returns them, investigating.)

ayon_server/graphql/resolvers/products.py
```diff
@@ -475,16 +476,11 @@ async def get_products(
...
-            term = term.replace("'", "''")
-            sub_conditions.append(f"products.name ILIKE '%{term}%'")
-            sub_conditions.append(f"products.product_type ILIKE '%{term}%'")
-            sub_conditions.append(f"folder_ex.path ILIKE '%{term}%'")
-
-            condition = " OR ".join(sub_conditions)
-            sql_conditions.append(f"({condition})")
+        if cond := build_search_conditions(
+            search,
+            ["products.name", "products.product_type", "hierarchy.path"],
```
filipvnencak:
```suggestion
            ["products.name", "products.product_type", "folder_ex.path"],
```

ayon_server/graphql/resolvers/versions.py
```diff
@@ -606,21 +607,12 @@ async def get_versions(
...
-
-            sub_conditions.append(f"products.name ILIKE '%{term}%'")
-            sub_conditions.append(f"products.product_type ILIKE '%{term}%'")
-            sub_conditions.append(f"folder_ex.path ILIKE '%{term}%'")
-
-            condition = " OR ".join(sub_conditions)
-            sql_conditions.append(f"({condition})")
+        if cond := build_search_conditions(
+            search,
+            ["products.name", "products.product_type", "hierarchy.path"],
```
filipvnencak:
```suggestion
            ["products.name", "products.product_type", "folder_ex.path"],
```

filipvnencak (changes requested):
The merge with develop didn't fully land. Products and versions now join `folder_ex`, but the search still references `hierarchy.path`. causing 500s, message:  missing FROM-clause entry for table "hierarchy"


  
Image

filipvnencak (approved):
Thanks, everything working 😉

kalisp:
There is new task_search in `folders` where it should be used too. Won't be merging to RC right now.

kalisp:
Converted new `task_search` on `folders` (Overview tab) to standardized function.

ayon_server/graphql/resolvers/common.py
```diff
@@ -404,3 +404,46 @@ def get_has_links_conds(
...
+    or no conditions could be built.
+    """
+    parts = search.split(",")
+    t1_conds = []
+
+    for part in parts:
+        terms = part.lower().replace("'", "''").split(" ")
+        t2_conds = []
+        for term in terms:
+            term = term.replace("\\", "\\\\").replace("_", "\\_").strip()
```
filipvnencak: % isn't escaped, so search: "%" matches every row, and "10%" matched unrelated tasks locally.

```suggestion
            term = term.strip()
            term = term.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
```

ayon_server/graphql/resolvers/common.py
```diff
@@ -404,3 +404,46 @@ def get_has_links_conds(
...
+    Within each term, the specified columns are OR'd together.
+
+    Returns a SQL condition string, or None if search is empty
+    or no conditions could be built.
+    """
+    parts = search.split(",")
+    t1_conds = []
+
+    for part in parts:
+        terms = part.lower().replace("'", "''").split(" ")
```
filipvnencak: `split(" ")` misses tabs/newlines. Bare `split()` also drops empty terms, so the `if not term` guard below becomes redundant.

```suggestion
        terms = part.lower().replace("'", "''").split()
```

ayon_server/graphql/resolvers/common.py
```diff
@@ -404,3 +404,46 @@ def get_has_links_conds(
...
+    search: str,
+    columns: list[str],
+    *,
+    version_check: bool = False,
+) -> str | None:
+    """Build SQL search conditions from a search string.
+
+    The search string is split by commas (OR between comma-separated parts).
+    Within each part, slugified terms are AND'd together.
+    Within each term, the specified columns are OR'd together.
```
filipvnencak: Docstring still describes the old behaviour — terms aren't slugified any more, `min_length` is gone, and `version_check` isn't mentioned.

```suggestion
    The search string is split by commas (OR between comma-separated parts).
    Within each part, whitespace-separated terms are AND'd together.
    Within each term, the specified columns are OR'd together.
    With version_check, a numeric term (or "v123") also matches versions.version.
```

ayon_server/graphql/resolvers/common.py
```diff
@@ -404,3 +404,46 @@ def get_has_links_conds(
...
+
+    Returns a SQL condition string, or None if search is empty
+    or no conditions could be built.
+    """
+    parts = search.split(",")
```
filipvnencak: Dropping slugify also dropped its separator set (`,./\;:!|*^#@~+-_=`), so punctuation no longer splits terms: on demo_Commercial `sh-010` now returns 0 folders while `sh 010` and `sh010` both return the `sh010` folder. If that's not intended, splitting each part on those characters would keep the old behaviour without reintroducing the ASCII transliteration you removed.

ayon_server/graphql/resolvers/common.py
```diff
@@ -404,3 +404,46 @@ def get_has_links_conds(
             f"{id_field} IN (SELECT input_id FROM project_{project_name}.links)",
         ]
     raise ValueError("Wrong has_links value")
+
+
+def build_search_conditions(
```
filipvnencak: The standardization stops at the GraphQL resolvers — `api/tasks_folders/tasks_folders.py:90`, `api/folders/search.py:123,189,241` and `graphql/resolvers/events.py:104` still use the old slugify approach. The Overview page sends the same search string to both the GraphQL resolvers and REST `tasksFolders`/`folders/search`, so the two paths will now disagree on non-ASCII and punctuation. If those get folded in later, this helper probably belongs in `ayon_server/utils` rather than under `graphql/resolvers`.

ayon_server/graphql/resolvers/tasks.py
```diff
@@ -471,21 +472,16 @@ async def get_tasks(
...
-            for term in terms:
-                t2_conds.append(
-                    f"(tasks.name ILIKE '%{term}%'"
-                    f"OR tasks.label ILIKE '%{term}%'"
-                    f"OR tasks.task_type ILIKE '%{term}%'"
-                    f"OR hierarchy.path ILIKE '%{term}%')"
-                )
-            t1_conds.append(SQLTool.conditions(t2_conds, "AND", add_where=False))
-        sql_conditions.append(SQLTool.conditions(t1_conds, "OR", add_where=False))
+        if cond := build_search_conditions(
```
filipvnencak, line 475: Dropping `min_length` is fine here, the frontend only fires the search on Enter, not per keystroke, so there's no 1-char scan on every key.

filipvnencak (changes requested):
Tested, everything works, only the % escaping looks like a real bug.

filipvnencak (approved):

PR #1054, opened 2026-08-10
Remove version list view

martastain:
This pull request refactors how the latest version of a product is queried and managed throughout the codebase. The changes remove reliance on the `version_list` materialized view and instead use direct queries on the `versions` table, improving both query efficiency and maintainability. Additionally, an index is updated to optimize these new queries.

**Database and Query Refactoring:**

* Replaced queries that used the `version_list` materialized view with direct queries on the `versions` table, selecting the latest version by ordering on the `version` field and limiting to one. This affects version resolution logic in both API and GraphQL layers. [
* Updated the SQL join in product resolvers to use a lateral join that aggregates version IDs and numbers directly from the `versions` table, removing dependency on the `version_list` view.

**Codebase Cleanup:**

* Removed the `refresh_views` class method from the `Version` entity, as refreshing the materialized view is no longer necessary.

**Note**
This PR does not remove version_list view from the older projects. It will be kept in order to allow downgrading. The view itself should be removed later

This PR significantly speeds up version saving

### DIFF
--- a/api/resolve/__init__.py
+++ b/api/resolve/__init__.py
@@ -161,8 +161,11 @@ def get_version_conditions(version_name: str | None) -> list[str]:
         return [
             """
             v.id in (
-                SELECT l.ids[array_upper(l.ids, 1)]
-                FROM version_list AS l
+                SELECT vv.id
+                FROM versions vv
+                WHERE vv.product_id = s.id
+                ORDER BY vv.version DESC
+                LIMIT 1
             )
         """
         ]
--- a/ayon_server/entities/version.py
+++ b/ayon_server/entities/version.py
@@ -86,17 +86,6 @@ class VersionEntity(ProjectLevelEntity):
                 self.task_id,
             )
 
-    @classmethod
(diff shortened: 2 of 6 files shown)


kalisp (approved):
Seems ok, tested publishes, still publishing.

Don't have test for create a sh.ton of version to actually measure performance gain though.

kalisp:
What was the idea about `version_list` at the beginning, just out of curiosity.

kalisp:
To have list of versions per product without querying? This PR is moving impact from creation of version to quering of versions. Is that actually good exchange? (As querying might be much more often than writing.) Or it shouldn't be too much of an impact with a index?

kalisp:
Adding context from Discord discussion: Updating materialized view slows down ingestion process for large number of versions (and large project) tremendously. One approach would be to refresh view only all versions are created, but it was deemed that view is not necessary anymore as it was replaced by featured versions in the product page.

Reading performance seems similar/same.

PR #1059, opened 2026-08-11
Fix nested SettingsModel defaults

BigRoy:
## Description of changes

Fix nested SettingsModels like:

```python
class WriteKnobsModel(SettingsModel):
    custom: list[KnobModel] = SettingsField(
        title="Additional Knobs",
        default=[
            KnobModel(
                type="text",
                name="mov64_codec",
                text="ap4h",
            ),
            KnobModel(
                type="boolean",
                name="mov64_write_timecode",
                boolean=True,
            )
        ]
    )

class IntermediateOutputModel(BaseSettingsModel):
    write_knobs: WriteKnobsModel = SettingsField(
        default_factory=WriteKnobsModel,
        title="Write Knobs",
        description=(
            "Configure knobs on the output write node"
        )
    )
    
    
class MainModel(BaseSettingsModel):
    outputs: list[IntermediateOutputModel] = SettingsField(
        default_factory=list,
        title="Baking streams"
    )
```

The defaults to `custom` key should automatically be set when adding a new entry to `outputs` in the Settings in frontend.

### Technical details

I feel like this 'fix' is way too convoluted and there must be something wrong - but just putting this as draft because it does resolve the issue here. However we should likely need to look at some other 'correct fix'.

### Additional context

Solves case described here: https://github.com/ynput/ayon-nuke/pull/344#discussion_r3758358228

This is not a high priority at all.

### DIFF
--- a/api/addons/project_settings.py
+++ b/api/addons/project_settings.py
@@ -20,7 +20,10 @@ from ayon_server.lib.postgres import Postgres
 from ayon_server.logging import logger
 from ayon_server.settings import BaseSettingsModel
 from ayon_server.settings.overrides import extract_overrides, list_overrides
-from ayon_server.settings.postprocess import postprocess_settings_schema
+from ayon_server.settings.postprocess import (
+    populate_schema_defaults,
+    postprocess_settings_schema,
+)
 from ayon_server.settings.set_addon_settings import set_addon_settings
 
 from .common import (
@@ -62,6 +65,7 @@ async def get_addon_project_settings_schema(
     }
 
     schema = copy.deepcopy(model.schema())
+    populate_schema_defaults(schema, model)
     await postprocess_settings_schema(schema, model, context=context)
     schema["title"] = addon.friendly_name
     return schema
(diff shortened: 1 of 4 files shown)


moonyuet:
It is working in my side with https://github.com/ynput/ayon-nuke/pull/344#discussion_r3758358228 through testing.

BigRoy:
This is just there to show that there's a bug that is solvable. The fix (created by AI) I'd say feels very off. I don't think we'd need to do this much mumbo jumbo and there must be an easier way.

What I really think the issue is is that in most integrations some BaseSettingModels may not define `default` for all attributes, e.g. a `str` attribute might not explicitly have `default=""` passed or alike. And maybe as we convert the schemas we need to stop deeper conversion because we can't initialize the base `MyModel()` to get to the defaults because you can't actually initialize it without passing the arguments that don't have defaults, perhaps? No idea. Anyway, it's not urgent, but good to be aware there is a bug here where defaults like in the Nuke example don't come through.

PR #1069, opened 2026-08-13
Enums: Added new endpoint for listing enums

### LINKED ISSUE #1062: Create endpoint to list available enum resolvers
### Problems

We have bunch of dynamic enum resolvers, but no easy way to list them for FE to choose from them.

### Proposal

Create endpoint which will return list of all dynamic enums.

kalisp:
## Description of changes

Adds enpoint `enums` (GET) to list all available enum resolvers.

### Technical details

Not completely sold on `enums` route, open for better names and with form of `settings_form`.  (But that could get updated when some resolvers will actually have `settings_form`.)

Contains also change of `get_accepted_params` to `AttributeType` for easier testing.


api/enums/__init__.py
```diff
@@ -61,3 +61,9 @@ async def get_enum(
         user=current_user,
         **context,
     )
+
+
+@router.get("/enums", response_model=list[EnumResolverInfo], tags=["Enums"])
```
martastain: This should be `/enum`  - as an entrypoint for `/enum/{enumName}`

ayon_server/enum/enum_registry.py
```diff
@@ -107,3 +133,21 @@ async def create_item(
             raise BadRequestException(f"Unknown enum resolver '{key}'")
 
         await resolver.create_item(item, project_name, **kwargs)
+
+    @classmethod
+    async def list_resolvers(cls) -> list[EnumResolverInfo]:
+        result = []
+        for name, resolver in cls.resolvers.items():
+            params = await resolver.get_accepted_params()
```
martastain, line 142: This is probably the last opportunity to change accepted params to `dict[str, AttributeType]` or `dict[str, str | int | float]`.  AttributeType is widely recognized in the codebase so it will be consistent - now we have `"str"` instead `"string"`.

martastain (changes requested):

api/enums/__init__.py
kalisp: Changed

ayon_server/enum/enum_registry.py
kalisp: Was saving it for separate PR but it would cause conflict, so I merged it all together.

ayon_server/enum/enum_registry.py
```diff
@@ -11,6 +12,31 @@
...
+    accepted_params: Annotated[
+        dict[str, AttributeType],
+        Field(
+            title="Accepted parameters",
+            description="Dictionary of accepted query parameters and their type names",
+            example={"project_name": "string", "include_root": "bool"},
+        ),
+    ]
+    settings_form: Annotated[
+        list[dict[str, Any]] | None,
```
filipvnencak: From the frontend side, list[dict[str, Any]] generates Record<string, any>[] in the OpenAPI client.Typing it as `list[SimpleFormField] | None `would give us the actual field schema (type/name/label/options) for free from codegen.

filipvnencak (changes requested):
Just one tiny issue, fix would help our codegen.

ayon_server/enum/enum_registry.py
kalisp: ok

filipvnencak (approved):

martastain (approved):
Works. Perfect

PR #1077, opened 2026-08-14
Versions: Defer querying the hasReviewables and latestComments for all versions

BigRoy:
## Description of changes

Versions resolver: Defer querying the hasReviewables and latestComments for all versions if not used for filtering, instead join them after the filtering and pagination applies so we can apply it over less rows

### Technical details

This _may_ improve performance in large projects when NOT filtering by hasReviewables or latestComments but they are requested in the output data. Note that the products table generally always requests hasReviewables, because it uses that data also for the little "play" icon for the thumbnail - as such, this may affect quite a few views, especially those that'd otherwise generate massive lists before pagination.

### Additional context

I didn't actually do any performance testing whether this is a noticeable benefit - just wanted to put it out there as a possibility.

### DIFF
--- a/ayon_server/graphql/resolvers/versions.py
+++ b/ayon_server/graphql/resolvers/versions.py
@@ -214,6 +214,8 @@ async def get_versions(
         "folder_ex.path AS _folder_path",
         "products.name AS _product_name",
     ]
+    deferred_joins = []
+    deferred_columns = []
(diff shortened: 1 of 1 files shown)


martastain:
been there, tried that. performance improvement is minimal, code readability suffers :-(

PR #1080, opened 2026-08-17
Column Stats: Use double quotes to keep column names format

kalisp:
## Description of changes

Column names with camelCase (`resolutionWidth`, `frameStart`) for column stats were returned lowercased in graphQL response. This resulted FE not matching values for them, therefore not displaying them.

### Technical details

Check `fps` and `width` columns, both should be showing distribution values.

image

Commit contains single quotes and also escaped double quotes, tried to standardize them, but commit pre-hook strips it back.

### DIFF
--- a/ayon_server/graphql/resolvers/field_stats.py
+++ b/ayon_server/graphql/resolvers/field_stats.py
@@ -84,30 +84,30 @@ def _get_stats_for_column(
     """Returns SQL fragments for statistics based on column data type."""
     if data_type in ("numeric", "int", "float"):
         return [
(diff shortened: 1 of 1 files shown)


martastain:
It does not work for me :-(

image

kalisp:
> It does not work for me :-(

Where exactly does this happen?

martastain:
This seems to do the trick, but i'm still having issues with certain attributes. but so far, it looks like a frontend issue
image

martastain (approved):

PR #1082, opened 2026-08-18
Permissions: Allow update HERO version with Restrict attribute update

### LINKED ISSUE #1068: YN-0993: hero version and attribute permissions

Modifying attribute permissions like 'clip in', 'clip out', 'handle start', 'handle end', 'start frame' and 'end frame' will exhibit permission errors during publish for Hero version if the user has restricted Attribute Update permissions, however due to the nature of hero versions it's very common for publishing from a user to have to update some attributes to it.

&nbsp;

To handle this case, we want to use the "Restrict publishing" as addition permissions case for updating HERO versions. If restrict publishing is NOT enabled for the user and they do have access to the version, then do allow updating attributes of the version if it is a hero version.

&nbsp;

When restrict publishing is not enabled for the user then ignore restrictions on "updating attributes" for hero versions. 

We may want to get more granular on the permissions model (e.g. separate attribute update permissions per entity type) but even that wouldn't solve it between hero version and regular version, so for now we deem this the best approach. This "attribute update" for hero versions hence is merely for "publishing" behavior, hence we use the Restrict publishing permissions for now.

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1536647516207058974)

_This issue was automatically created from AYON ticket [YN-0993]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=79eeb98325b946b0b52bf2f4937a68c1)

kalisp:
## Description of changes

This handle situation where `Restrict attribute update` is enabled. This offer set of attributes (`fps`, `resolutionWidth`) and fields (`product_name`, `product_type`) that could be updated by the artists (or that role) ONLY.

Without this PR HERO version couldn't be updated as it needs updates to `version` - points to latest version number, `files` - contains updated information about file(s) itself.

### Technical details

This is simplest solution to issue use case. I am not too fussy about `version`, that could be set as always updatable field (don't see use case where we would like to control this more granularly - would we want to create/update HERO versions only to selected group of artist (role)??

`files` is a bit more dangerous, but we offer granular control on that field. Question is if it shouldn't be added as additional toggle for `Restrict attribute update`.

I don't think that using `"Restrict publishing"` from issue description is too relevant. That Setting controls if artist could create/update in particular folder (eg. context). That check is resolved before, either I can or cannot create version there, it will not bring any additional granularity for check which attributes or fields should be updated for which version.

Adding new values to class variable seems a bit tight, we could create methods for each entity to override if we would like to have more control.


BigRoy:
I think it's not the right approach - we're just dropping our level of security, right? I feel like we should really focus on getting the permissions to only affect the hero version like the issue describes. Let's discuss.

We should definitely refrain from anyone to touch "version" on EVERY version. If this were to only apply to hero versions, then it's ok. If at least it's still limited if they are restricted from publishing.

kalisp:
Added 2 additional methods to BaseEntity for each entitty to provide list of always writable.

Not completely happy with return from `RepresentationEntity.always_writable_fields` but as `patch` is not async, I don't know how to check if parent version is hero.

ayon_server/entities/version.py
```diff
@@ -110,6 +113,11 @@ async def ensure_update_access(self, user, **kwargs) -> None:
             "publish",
         )
 
+    def always_writable_fields(self, user: Optional["UserEntity"] = None) -> list[str]:
+        if self.version < 0:
+            return ["version", "data"]
+        return []
```
BigRoy: Why do these take the `user` argument that you're never passing? 🤔

ayon_server/entities/core/base.py
```diff
@@ -80,10 +80,15 @@ def patch(self, patch_data: BaseModel, user: Optional["UserEntity"] = None) -> N
...
-                        perms.attrib_write.attributes + ALWAYS_WRITABLE_ATTRS
+                        perms.attrib_write.attributes
+                        + ALWAYS_WRITABLE_ATTRS
+                        + self.always_writable_attrs()
+                    )
+                    writable_fields = (
+                        perms.attrib_write.fields
+                        + ALWAYS_WRITABLE_FIELDS
+                        + self.always_writable_fields()
                     )
```
BigRoy, line 84: write attrs and writable fields should really be `set[str]` here - it'd avoid duplicated and would make the `if attr not in writable_attrs` logic below much faster?

ayon_server/entities/representation.py
```diff
@@ -75,6 +78,9 @@ async def ensure_update_access(self, user, **kwargs) -> None:
             "publish",
         )
 
+    def always_writable_fields(self, user: Optional["UserEntity"] = None) -> list[str]:
+        return ["files", "data"]
```
BigRoy: Can we make this more secure somehow with making things even slower? Like, do we technically have access here to know whether it's a hero version and then return this?

ayon_server/entities/version.py
```diff
@@ -110,6 +113,11 @@ async def ensure_update_access(self, user, **kwargs) -> None:
             "publish",
         )
 
+    def always_writable_fields(self, user: Optional["UserEntity"] = None) -> list[str]:
+        if self.version < 0:
```
BigRoy:
```suggestion
        # Allow editing for hero versions because those are
        # published in-place of an existing version
        if self.version < 0:
```

ayon_server/entities/representation.py
BigRoy: @martastain let's discuss this next week.

BigRoy: @kalisp I suspect we'll also need to give access to `attrib` in general because the user may update frame start, frame end, etc. and other attributes on the version update. As such, I believe both this case and the version case will need to allow the attribute access as well.

May be good to do some test runs with that as well, where we change version resolution, frame ranges, handles and perhaps FPS between the publishes, etc.

ayon_server/entities/version.py
kalisp: Passed.

kalisp:
I added guard on Representation to check if it belongs to hero version.

It brings a question if `always_writable_fields` and `always_writable_attrs` are really necessary and if that check of updatable attributes/fields should be completely skipped for `HERO` version/representation. 

It could be implemented by replacing these two methods with `skip_writable_check` which would contain current implementation of `always_writable_fields` (eg check if version/representation is HERO).

ayon_server/entities/representation.py
```diff
@@ -49,6 +52,7 @@ def preprocess_record(record: dict[str, Any]) -> dict[str, Any]:
             vname = version_name(record["version"])
             rname = record["name"]
             record["path"] = f"/{hierarchy_path}/{product_name}/{vname}/{rname}"
+        record["belongs_to_hero"] = record["version"] < 0
```
BigRoy, line 52: Wouldn't this affect the data we're patching/setting? 🤔 

Can't we just access `self.payload.version` down below?

kalisp: There is no `self.payload.version`, there is only `self.payload.version_id` which doesn't differentiate between regular/HERO.

`preprocess_record` has access to fields from BASE_GET_QUERY where I have access to `version` value itself.

BigRoy: Not sure if there's much reason to put `belongs_to_hero` there only instead of just putting `version` in there but I suppose there are reasons. Fine with me.

BigRoy (changes requested):
This does not give the user enough permissions.

```
ERROR: !!! Creating of hero version failed. Previous hero version maybe lost some data!
Traceback (most recent call last):
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2608251347_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\User\AppData\Local\Ynput\AYON\addons\core_1.9.10+dev\ayon_core\plugins\publish\integrate_hero_version.py", line 133, in process
    self.integrate_instance(
  File "C:\Users\User\AppData\Local\Ynput\AYON\addons\core_1.9.10+dev\ayon_core\plugins\publish\integrate_hero_version.py", line 527, in integrate_instance
    op_session.commit()
  File "C:\roydev\dev\ayon-launcher\build\AYON 1.6.3\dependencies\ayon_api\operations.py", line 837, in commit
    self._con.send_background_batch_operations(
  File "C:\roydev\dev\ayon-launcher\build\AYON 1.6.3\dependencies\ayon_api\server_api.py", line 2578, in send_background_batch_operations
    self._validate_operations_result(
  File "C:\roydev\dev\ayon-launcher\build\AYON 1.6.3\dependencies\ayon_api\server_api.py", line 2682, in _validate_operations_result
    raise FailedOperations(
ayon_api.exceptions.FailedOperations: Operation "95f547d2-58c2-42f2-bc54-7c5580c2c783" failed with data:
{
    "id": "95f547d2-58c2-42f2-bc54-7c5580c2c783",
    "type": "update",
    "entityType": "version",
    "entityId": "8e7ac96aa0b011f1b8c360cf84748747",
    "data": {
        "version": -113,
        "data": {
            "time": "20260825T201649Z",
            "author": "johny",
            "ayon_app_name": "fusion/20",
            "host_name": "fusion"
        },
        "attrib": {
            "source": "C:\\projects\\test\\apps\\fusion\\work\\compositing\\tes_fusion_compositing_v113_tests.comp",
            "frameEnd": 15,
            "filename": "tes_fusion_renderCompositingLong_v0113"
        }
    }
}
Detail: You are not allowed to modify frameEnd attribute in test.
```
I must be able to adjust attributes on hero versions as well.
This seems to confirm the case I described before here: https://github.com/ynput/ayon-backend/pull/1082#discussion_r3811921582

I also suspect the user must be allowed to change `traits` as well - and with [that potentially `tags` (because it does support `"versionTags"` in `instance.data`. ](https://github.com/ynput/ayon-core/blob/59cf2bfd16488f0db9010498d5b2836645385163/client/ayon_core/plugins/publish/integrate_hero_version.py#L246)

kalisp:
> This does not give the user enough permissions.

So this bring up the route with just skipping that check altogether, without 2 `allowed` methods. Ok, will change it.

ayon_server/entities/representation.py
kalisp: Replaced allow list with single check method

kalisp:
Replaced two `allow_writable*` methods which were unnecessary granular with single check if entities are/belong to HERO.

ayon_server/entities/core/base.py
```diff
@@ -199,3 +198,13 @@ def updated_at(self) -> float:
...
+
+    def _skip_check(self) -> bool:
+        """Checks if current entity is not HERO version or repre."""
+        if hasattr(self.payload, "version"):
+            return self.payload.version < 0
+
+        if hasattr(self.payload, "belongs_to_hero"):
+            return self.payload.belongs_to_hero
+
+        return False
```
BigRoy: Should this be more explicit.

Return `False` on the base class.
Return `if version < 0` on the version class
Return `if belongs_to_hero` on the representation class.

It feels a bit dirty to be here on the base class - since it's not really base class logic.

Also `_skip_check`, we may want to make that more explicitly named, `_skip_patch_permissions_check`?

@martastain thoughts?

BigRoy (approved):
This does now give a user with the limited permissions the right permissions needed to publish the hero verisons (and adjust them).

There's some cosmetic notes that I'll leave to @martastain though to confirm whether he wants it cleaned.

ayon_server/entities/core/base.py
martastain: Agreed. Those are valid points.

kalisp: Done

BigRoy: Should we still rename the method to `_skip_patch_permissions_check`? Also, it doesn't need to be public right - because we're always calling it on `self`?

kalisp: Overriding private method seems wrong.

BigRoy: Fine with me - can we then do `skip_patch_permissions_check`?

kalisp: Ok, done.

ayon_server/entities/version.py
```diff
@@ -170,3 +170,8 @@ def author(self) -> str:
     @property
     def path(self) -> str:
         return self._payload.path  # type: ignore
+
+    def skip_patch_permissions_check(self) -> bool:
+        """Checks if current entity is not HERO version or repre."""
```
BigRoy:
```suggestion
        """Checks if current entity is HERO version."""
```

ayon_server/entities/representation.py
```diff
@@ -90,3 +91,7 @@ def version_id(self, value: str):
     @property
     def parent_id(self) -> str:
         return self.version_id
+
+    def skip_patch_permissions_check(self) -> bool:
+        """Checks if current entity is not HERO version or repre."""
```
BigRoy:
```suggestion
        """Checks if current entity is HERO version representation."""
```

ayon_server/entities/core/base.py
```diff
@@ -199,3 +198,8 @@ def updated_at(self) -> float:
     @updated_at.setter
     def updated_at(self, value: float) -> None:
         self._payload.updated_at = value  # type: ignore
+
+    def skip_patch_permissions_check(self) -> bool:
+        """Checks if current entity is not HERO version or repre."""
```
BigRoy:
```suggestion
        """Return whether patch permission check should be skipped for this entity."""
```

BigRoy (approved):
LGTM, just some cosmetics.

BigRoy (approved):
👍

PR #1085, opened 2026-08-19
Hide internal folder from the API

jakubfiala:
## PR Checklist

-   [ ] Referenced issue is linked

## Description of changes

Introduces the concept of an internal folder called `__ayon_internal__` which can be used to store a hierarchy of entities meant to be hidden in regular AYON UI. This folder and its descendants are never returned by the API.

When creating a new folder or patching an existing one, the folder name cannot start with `__ayon_internal__`, otherwise a Bad Request exception is thrown.

### Additional context

This functionality is useful for whenever we want to create a hierarchy of entities that isn't supposed to appear in the AYON UI. For instance, an addon might keep its own folder with various entities which are only supposed to be accessed by it.

### DIFF
--- a/api/folders/folders.py
+++ b/api/folders/folders.py
@@ -3,6 +3,8 @@ from fastapi import Query
 from ayon_server.api.dependencies import CurrentUser, FolderID, ProjectName
 from ayon_server.api.responses import EmptyResponse, EntityIdResponse
 from ayon_server.entities import FolderEntity
+from ayon_server.exceptions import BadRequestException
+from ayon_server.helpers.hierarchy_cache import AYON_INTERNAL_FOLDER_NAME
 from ayon_server.operations.project_level import ProjectLevelOperations
 
 from .router import router
@@ -38,6 +40,9 @@ async def create_folder(
 ) -> EntityIdResponse:
     """Create a new folder."""
 
+    if post_data.name.startswith(AYON_INTERNAL_FOLDER_NAME):
+        raise BadRequestException("reserved folder name")
+
     ops = ProjectLevelOperations(project_name, user=user)
 
     ops.create("folder", **post_data.dict(exclude_unset=True))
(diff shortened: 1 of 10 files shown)


jakubfiala:
@martastain this is now ready for review. I added the internal folder filtering to various GraphQL resolvers with a flag to skip it (defaults to False).

kalisp:
Wouldn't it be better to specify new `folder_type` (probably not exposing it in Anatomy as it could get removed by some TD for clean up reasons). There is already index on `folders.folder_type`.

(Not sure if it wouldn't causing issues in `folder_types` table and how they are handled there. There is nothing there that would differentiate 'regular' and 'special' folder types. (For FE to show/hide specific ones in Anatomy.)

(Just spit-balling here. I'll let Martin to properly respond.)

(And maybe more description about intended use case would be good, for future posterity. As changed lines could get traversed back to PR which is usually latest place where to glance additional information that usually won't get into any documentation.)

jakubfiala:
@kalisp i mean, if you think making a new folder type is better, we can do it that way. I'm not sure in what way it's better though. I don't envision this being used a lot and having lots of different internal folders.

Good point re: describing the use case, i'll do that now.

martastain (approved):

martastain:
> Wouldn't it be better to specify new folder_type (probably not exposing it in Anatomy as it could get removed by some TD for clean up reasons). There is already index on folders.folder_type.

Not really. Keep in mind that you need to filter out children as well (so for example versions in internal folder's subfolder won't appear in the default graphql query - this is much easier to do using a folder name, which we can rely on is the first element of entity path.

BigRoy (commented):
⚠️ This PR made it so that root folders are now not returned from graphQL. 

@martastain is investigating and may make a bugfix.

martastain:
Fixed by https://github.com/ynput/ayon-backend/commit/5020def65251d5cbc84ed49643f5757c67fb3e9f

BigRoy:
> Fixed by [5020def](https://github.com/ynput/ayon-backend/commit/5020def65251d5cbc84ed49643f5757c67fb3e9f)

@martastain that commit seems to affect only one place, but looking at this PR a similar section of code is actually in multiple places? Are the others potentially still wrong?

martastain:
@BigRoy I've tested other resolvers as well as hierarchy caching mechanism and it worked for me. i think it really affected root folders because of that left join.

PR #1087, opened 2026-08-21
Guests: Added guest user patch endpoint

### LINKED ISSUE #722: Guests: PATCH guest user endpoint
Guest users have `fullName` attribute and this should be updatable without having to add the user again.

The guest model might expand in the future with more attributes so a general PATCH approach might be best.

kalisp:
## Description of changes

Added new PATCH endpoint for admins to update attributes of guest accounts in `project.data`.

### Additional context

image

### DIFF
--- a/api/projects/guests.py
+++ b/api/projects/guests.py
@@ -34,6 +34,40 @@ class GuestUserModel(OPModel):
     ] = "pending"
 
 
+class GuestUserPatchModel(OPModel):
+    new_email: Annotated[
+        str | None,
+        Field(
+            title="New Email",
+            example="foo.bar@example.com",
+        ),
+    ]
+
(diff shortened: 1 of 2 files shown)


kalisp:
Not sure if we want to completely remove any stored fields. Currently it is not really possible as it skips unset fields.

Innders (commented):
Just found out that the guest endpoints aren't used in the frontend so I have no way of testing it.

martastain:
Semantically, i believe the field shoud be named `email` not `new_email`. I did the same mistake before.

kalisp:
> named `email` not `new_email`

Renamed to `email` in payload, passing it as dictionary instead of kwargs.

martastain (approved):

PR #1096, opened 2026-08-27
Add flag to suppress notifications when saving entities, creating reviewables or deleting activities

jakubfiala:
## PR Checklist

-   [ ] Referenced issue is linked

## Description of changes

Adds the ability to perform certain operations without sending websocket notifications. The operations are:

- `EntityList.save`
- `create_reviewable`
- `delete_activity`

These functions get an optional `create_events` boolean flag which is true by default.

### Additional context

For some addons, we have the use case of frequently updating an entity list of versions and replacing reviewables on those versions. This allows the addon to auto-save user's changes on the list/version. In order not to flood the event logs and reports with too many messages, we need a way to make the updates without sending notifications.

### DIFF
--- a/ayon_server/activities/create_activity.py
+++ b/ayon_server/activities/create_activity.py
@@ -58,7 +58,7 @@ async def create_activity(
     sender: str | None = None,
     sender_type: str | None = None,
     bump_entity_updated_at: bool = False,
-    send_notifications: bool = True,
+    create_events: bool = True,
 ) -> str:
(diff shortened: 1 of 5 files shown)


martastain:
we don't use term notifications anywhere iirc. i'd use create_events instead.

jakubfiala:
@martastain good point, will do

jakubfiala:
done!

PR #1097, opened 2026-08-27
Settings: Added new endpoint to migrate single addon settings between two versions

### LINKED ISSUE #533: AY-7582_add copy project overides to addon copy settings
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

when not using general copy bundle settings when a new bundle is created (on purpose) and I want to manually copy settings of individual addons, there is no way to include project overrides. These are only copied with bulk bundle copy.

We should have a possibility to copy project overrides on individual addons

### How would you imagine the implementation of the enhancemenent?
_No response_

### Describe alternatives you've considered:
_No response_

### Additional context:
link to discussion on [Discord](None)

_This issue was automatically created from Clickup ticket [AY-7582](https://app.clickup.com/t/86c2g4n9g)_

kalisp:
## Description of changes

Creates new endpoint `api/addons/{addon_name}/migrate` to manually trigger Settings migration for selected addon only. It contains `with_projects` flag to copy over also all `Project Overrides`.

When new bundle is promoted into production, settings for all addons are migrated between two bundles. This is more granular approach per addon and with limited use case.

### Additional context

Could be tested with this POST:

image

```
{
  "source_version": "1.2.6+dev",
  "target_version": "1.3.1+dev",
  "source_variant": "staging",
  "target_variant": "production",
  "with_projects": true
}
```


martastain (approved):
LGTM.... But let's merge it when the frontend part is ready

PR #1099, opened 2026-08-27
Implemented saving data.mediaInfo for video attached in comments

### LINKED ISSUE #1086: Files: extract media info for uploaded video files
## Story

Uploaded videos in comments have no media metadata, `/files/{id}/info` returns only size, filename and content type. We now play them with ayon-player (ayon-frontend#2119) and it needs fps. Frontend can't get fps from the browser so it guesses 24, frame numbers are wrong for 25/30 fps videos.

### Expected behavior

ffprobe at upload for `video/*` files, result stored and returned in `/files/{id}/info` and the activity `files` payload, same shape as reviewables `mediaInfo`.

kalisp:
## Description of changes

Uses `ffprobe` to store `mediaInfo` for videos uploaded via Activity comment in similar fashion as `reviewables` have.

Will be used for proper playback by correct `fps`.

### Additional context

Use `api/projects/ayon_project/files/{video_file_attached_in_comment}/info` for testing

### DIFF
--- a/api/files/files.py
+++ b/api/files/files.py
@@ -103,6 +103,23 @@ async def upload_project_file(
             file_id,
         )
 
+    if content_type.startswith("video"):
+        try:
+            media_info = await storage.extract_media_info(file_id)
+        except Exception as e:
+            logger.warning(f"Failed to extract media info for {file_id}: {e}")
+            media_info = {}
+        if media_info:
+            await Postgres.execute(
+                f"""
+                UPDATE project_{project_name}.files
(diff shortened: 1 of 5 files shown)


martastain (approved):

filipvnencak (changes requested):
Confirmed working. But can you add it to graphql and not rest? Data is already in activity data.files so it should be easy, and we don't need another api request.

kalisp:
image

filipvnencak (approved):
Tested working

PR #1101, opened 2026-08-31
Make default link types required

### LINKED ISSUE #144: Reserved link types
## Story

We'll need to reserve some basic built-in link types that we are using all across the board in pipeline code. 

Namely:

- `reference|version|version`
- `generative|version|version`
- `breakdown|folder|folder`
- `workflow|task|task`

these four have special pusposes and we should be able to rely on them heavily in various automation.

## Proposed solution

The easiest solution would be just to create a validator for the link types field of the Anatomy model. But from the UX perspective, we should also allow setting certain items in list[Submodel] fields as read only/non-removable.

martastain:
This pull request introduces logic to ensure that all required default link types are present when loading or deploying a project. The main focus is on improving data integrity by automatically adding any missing default link types to a project's configuration, both during project loading and deployment. The changes also improve logging for better traceability.

**Ensuring presence of default link types:**

* Added a new function `ensure_required_project_link_types` in `project.py` to check and create any missing default link types for a project during the loading process. This function is called in the `_load` method of `ProjectEntity` to guarantee required link types exist.
* Updated `deploy_project.py` to iterate through `default_link_types` and append any missing ones to the project's anatomy during deployment, with debug logging for visibility.

### DIFF
--- a/ayon_server/entities/models/submodels.py
+++ b/ayon_server/entities/models/submodels.py
@@ -47,3 +47,14 @@ class LinkTypeModel(OPModel):
         default_factory=dict,
         description="Additional link type data",
     )
+
+    def _identity(self) -> tuple[str, str, str]:
+        return (self.link_type, self.input_type, self.output_type)
+
+    def __hash__(self) -> int:
+        return hash(self._identity())
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, LinkTypeModel):
+            return NotImplemented
+        return self._identity() == other._identity()
--- a/ayon_server/entities/project.py
+++ b/ayon_server/entities/project.py
@@ -4,7 +4,7 @@ along with the project data, this entity also handles
 folder_types of the project and the folder hierarchy.
 """
 
-from collections.abc import Sequence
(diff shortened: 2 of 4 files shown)


kalisp:
How this could be actually tested? Trying to restore dump with missing link types?

martastain:
You could just delete some link types in develop and then switch to this branch... it should be repopulated :crossed_fingers:

BigRoy:
When should this trigger and recreate them?

I I delete Link Types from anatomy from a project, then restarting server - it doesn't recreate them.

Doesn't seem to work for me @martastain

martastain:
Fixed the check when cache is used

martastain:
> Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.

:middle_finger:

BigRoy (approved):
It seems to work now - whenever I try to remove the existing links, it seems to save with success, but upon refresh they are back. That does make it a bit confusing possibly, because it doesn't tell you in frontend that you can't seem to delete them?

Which makes me think, should we also show these entries differently in anatomy? And, could we?

These are the links it seems to enforce:
image

PR #1102, opened 2026-08-31
Auto-remove addons from bundle if they were removed

martastain:
This pull request adds validation to the handling of addons in the `update_bundle` function. Now, before updating a bundle, the code checks that each addon listed actually exists in the `AddonLibrary`. If an addon does not exist, it is removed from the bundle and a warning is logged. This helps prevent invalid or non-existent addons from being saved in the bundle.

(And most importantly - it gives back the ability to archive/change bundles, when they contain an addon that was removed, because originally, this triggered an error during bundle validation)


BigRoy (commented):
Code makes sense - @kalisp can you review it?

kalisp:
How is this supposed to be working?

Removed addon (server) (without restart), I can still select him in the Bundle addon version dropdown (which is I guess expected). But when I select it, it seems it gets saved ok.

I created new copy of bundle, selected that missing addon and saved it fine. (But updated method is probably not called, but `create_new_bundle` is called instead.)

(After server restart addon version gets removed in the bundle. If that is new expected behavior, then it work :).)

PR #1111, opened 2026-09-03
Fix projects resolver pagination with ACL

### LINKED ISSUE #1109: GraphQL: Project pagination is broken
In the projects resolver, project access is evaluated on node level, that skips projects the user don't have access to, but hasNextPage does not reflect it - if page size is lower than project count the user has access to, hasNextPage is not set to true and the rest of the projects is never loaded.

Suggested solution: Implement ACL check on SQL level instead of skipping projects later

martastain:
This pull request enhances the access control logic for the `get_projects` GraphQL resolver by restricting non-manager users to see only projects they have access to, based on their access groups. The main changes introduce a Common Table Expression (CTE) and JOIN to filter projects for non-manager users.

Previously, access control was implemented at the end of the chain, and when more than a half of projects were skipped cursor never reached the projects at the end of the page the user had access to

### DIFF
--- a/ayon_server/graphql/resolvers/projects.py
+++ b/ayon_server/graphql/resolvers/projects.py
@@ -78,7 +78,16 @@ async def get_projects(
     if fields.has_any("data", "bundle") or info.context["user"].is_guest:
         cols.append("data")
 
-    if not user.is_manager:
+    if user.is_guest:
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Code looks ok, ACL works, I tried to enforce pagination and it seemed to work.

PR #1113, opened 2026-09-08
Visibility filtering

martastain:
## Summary

This PR introduces a consistent **visibility filter** across the GraphQL API, letting clients ask for only visible entities, only hidden (inactive) entities, or everything.

A new `EntityVisibility` enum (`ALL`, `VISIBLE`, `HIDDEN`) is added as a `visibility` argument on the main list queries for folders, tasks, products, versions, representations, workfiles, and projects. Previously, "active" filtering was inconsistent between resolvers (some ignored it, some only checked the entity's own `active` flag); this PR makes visibility inheritance explicit and uniform.

## Expected behavior (visibility inheritance)

An entity is **visible** only if it and everything it inherits visibility from are active. Concretely:

* **Project** — only its own active state.
* **Folder** — its own active state *and* all ancestor folders active.
* **Task** — its own active state *and* all ancestor folders (its folder and that folder's ancestors) active.
* **Product** — its own active state *and* all ancestor folders active.
* **Version** — its own active state, its product's active state, *and* all ancestor folders active.
* **Representation** — its own active state, its version's active state, its product's active state, *and* all ancestor folders active.
* **Workfile** — its own active state, its task's active state, *and* all ancestor folders active.

**Note:** Products, versions, and representations do **not** take the *task*'s active state into account, even when a version is linked to a task. This is intentional:
- Products aren't linked to a task at all, only versions are.
- Versions don't require a task link either, so some versions could never be hidden by task state even if all tasks were deactivated.
- The task → version relationship is treated as metadata/tracking, not as something that should drive visibility - archiving a task shouldn't suddenly hide products/versions/representations that live under the same folder.

Workfiles are the exception, since they are directly owned by a task, so a workfile's visibility does depend on its task's active state.

`HIDDEN` is the logical negation of `VISIBLE` (i.e. "not everything in the chain above is active"), not simply "this entity's own `active` flag is false".

## Other changes

- SQL conditions for folders, kanban, and other resolvers were adjusted to correctly combine the entity's own state with its inherited/ancestor state (including edge cases like root folders, which have no parent).
- Internal (`__ayon_internal__`) folders and their descendants continue to be excluded from results unless explicitly requested via `includeInternalFolder`, now matched by exact path segment rather than a `LIKE` pattern.
- Visibility and internal-folder filtering are only applied when no specific `ids` are requested, so explicitly requesting an entity by ID still returns it regardless of visibility.

## Testing notes

Run `ayon setup` to apply the required database migration (adds an `active` column to `exported_attributes` and backfills inherited visibility)

### DIFF
--- a/ayon_server/graphql/resolvers/representations.py
+++ b/ayon_server/graphql/resolvers/representations.py
@@ -26,7 +26,7 @@ from ayon_server.sqlfilter import QueryFilter, build_filter
 from ayon_server.types import validate_name_list, validate_status_list
 from ayon_server.utils import SQLTool
 
-from .common import build_search_conditions
+from .common import ARGVisibility, EntityVisibility, build_search_conditions
 
 
 async def get_representations(
@@ -50,6 +50,7 @@ async def get_representations(
     search: Annotated[str | None, argdesc("Fuzzy text search filter")] = None,
     filter: Annotated[str | None, argdesc("Filter tasks using QueryFilter")] = None,
     include_internal_folder: ARGIncludeInternalFolder = False,
+    visibility: ARGVisibility = EntityVisibility.ALL,
 ) -> RepresentationsConnection:
     """Return a list of representations."""
 
@@ -73,6 +74,32 @@ async def get_representations(
         if not ids:
             return RepresentationsConnection()
         sql_conditions.append(f"representations.id IN {SQLTool.id_array(ids)}")
+    else:
+        if visibility == EntityVisibility.VISIBLE:
+            sql_conditions.append(
+                """
+                (
+                    representations.active
+                    AND versions.active
+                    AND products.active
+                    AND f_ex.active
+                )
+                """
+            )
+        elif visibility == EntityVisibility.HIDDEN:
+            sql_conditions.append(
+                """
+                (
+                    NOT representations.active
+                    OR NOT versions.active
+                    OR NOT products.active
+                    OR NOT f_ex.active
+                )
+                """
+            )
+
+        if not include_internal_folder:
+            sql_conditions.append(f"f_ex.path NOT LIKE '{AYON_INTERNAL_FOLDER_NAME}%'")
 
     if version_ids is not None:
         if not version_ids:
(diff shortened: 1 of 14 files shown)


ayon_server/graphql/resolvers/representations.py
```diff
@@ -73,6 +74,34 @@ async def get_representations(
...
+                    AND products.active
+                    AND f_ex.active
+                )
+                """
+            )
+        elif visibility == EntityVisibility.HIDDEN:
+            sql_conditions.append(
+                """
+                (
+                    NOT representations.active
```
kalisp, line 93: Shouldn't this be also on task|folder? 

If I have inactive tasks|folder, I would be expecting everything underneath to be inactive. 

Now I have inactive task where I published new version and its representations are not returned when `EntityVisibility.HIDDEN`

~~This would apply for `products`, `versions`, `tasks` (inactive folder).~~ It seems it look ok-ish in those. Or better, how I would expect it.

BigRoy: Just FYI, I think we agreed in a call with Martin, that Task would not hide products/versions/representations, only the folder-level would.

As such, nothing hides based on tasks' active state except for itself.

kalisp: Oook, could we add a comment for future posterity.

I don't see reason why it would behave differently, but there must be a use case where it does.

BigRoy: It's because products aren't linked to a task, but versions are. And it'd be odd if in a folder you'd suddenly be missing versions just because a task was archived. Versions are not required to be linked to a task too, as such - some of them could never hide even if deactivating all tasks.

The task -> version metadata we feel is more of a metadata tracking than the thing that defines whether it's visible or not.

kalisp (approved):
Tested, looks good.

ayon_server/graphql/resolvers/representations.py
BigRoy: @martastain can you put the expected behavior for this inheritance in the PR description?

E.g.

> * Project: only self active state
> * Folder: self active and all ancestor active
> * Task: self active and all folder ancestor active
> * Product: self active and all folder ancestor active
> * Version: self active, product active and all folder ancestor active
> * Representation: self active, version active, product active and all folder ancestor active
> 
> Note that product, version and representation do not rely for visibility on the _task_ active state - even if the version is linked to the task.

PR #1115, opened 2026-09-09
Added filter for repre extensions to versions graphql

### LINKED ISSUE #1114: YN-0926: Add file type filters for viewables in Loader

Add filter by extension in Loader so it is easy to find files with specific extension like  pdf, .mov, .mxf, .jpg, .jpeg, .png, etc.
 

To be able to list all versions in the browser (in desktop tools, similar to web browser) that are of a certain file type we'll need to be able to filter by representation.

&nbsp;

Preferably we can filter by extension, like e.g. matching with regex against the 'representation.attrib.path' - or another approach which could work acceptably well is to filter by 'representation.name' (because in majority of cases it matches the extension)

&nbsp;

In short, add a filter endpoint to the versions resolver (and with that likely also to the group/measuring and the products resolver) to be able to filter by representation name or any of the representation attributes (to allow matching against 'path').

&nbsp;

It's important that if no representation filter is applied, there's no big overhead to existing queries. So it shouldn't slow down the resolver.

And side note, the filter would match if it has ANY representation of that kind.

### Additional context:

_This issue was automatically created from AYON ticket [YN-0926]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=4ddb852872a244aaa7200a986aa1e8e6)

kalisp:
## Description of changes

Adds additional `representation_extensions` filter to search for versions having representation of particular extension.

This should be used for Loader and desktop tools to locate appropriate versions faster.

### Additional context

This could be tested only via calling `get_versions` via Postman or similar

### DIFF
--- a/ayon_server/graphql/resolvers/versions.py
+++ b/ayon_server/graphql/resolvers/versions.py
@@ -214,17 +214,6 @@ def available_joins(project_name: str) -> dict[str, str]:
                 LIMIT 1
             ) hv ON true
             """,
-        "repr": f"""
-            INNER JOIN LATERAL (
-                SELECT
-                    versions.id AS id,
-                    r.name as repr_name,
-                    COALESCE(r.attrib->>'path', '') AS repr_path
-                FROM project_{project_name}.representations r
-                WHERE r.version_id = versions.id
-                LIMIT 1
-            ) repr ON true
-        """,
     }
 
 
@@ -326,13 +315,6 @@ async def get_versions(
         list[str] | None,
         argdesc("List of tags to filter by"),
     ] = None,
-    representation_extensions: Annotated[
-        list[str] | None,
-        argdesc(
-            "List of file extensions (e.g. pdf, mov, .jpg) to filter by "
-            "representation path/name"
-        ),
-    ] = None,
     product_ids: Annotated[
         list[str] | None,
         argdesc("List of parent products IDs"),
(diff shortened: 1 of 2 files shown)


martastain (changes requested):
This feels really random. when we're joining with representations, we should probably have full representation_filter (so we're consistent with folder_ version_ task_ filters we already have. 

To match by multiple extensions, we could add a new operator to sqlfilter (my hot take `re` , that would resolve to "SIMILAR TO" postgres regex matcher - it will be slow anyways, since end of a string is not really indexable.

BigRoy:
Agree with Martin here - should standardize here.

kalisp:
Replaced with smth more similar to `product_filter` etc.

BigRoy (approved):
Amazing - this seems to work ❤️

ayon_server/graphql/resolvers/versions.py
```diff
@@ -45,6 +46,8 @@
 )
 from .sorting import get_attrib_sort_case, get_status_sort_case
 
+REPRESENTATION_EXTENSION_REGEX = re.compile(r"^[a-z0-9][a-z0-9._+-]*$")
```
martastain: This is now unused (as well as _normalize_representation_extensions)

martastain (changes requested):
Works great! Just get rid of that unused code and merge.

martastain (approved):

PR #1116, opened 2026-09-10
Simple form enhancements: conditionals, templating, enumResolver

martastain:
## Summary
- `ayon_server/forms/simple_form.py`: adds conditional `rules` (reusing `ayon_server.sqlfilter`'s `QueryCondition`/`QueryFilter` model) that let a field's `readOnly`/`disabled`/`hidden`/`value`/`options`/... be patched based on the current value of an earlier field in the same form.
- Adds `enumResolver`/`enumResolverParams` to `select`/`multiselect` fields, so options can be fetched live from `/api/enum/{name}` instead of being baked into the form; params may reference an earlier field's value via `{{fieldName}}` templating.
- Forward-reference guard: a rule's `when` or a resolver's `enumResolverParams` may only reference fields already added earlier in the same form - enforced at construction time, not just documented.
- `FormOptionItem` replaces the old `FormSelectOption` as the canonical (sparse - only `value` required) option shape; `FormSelectOption` kept as a backwards-compatible alias. `EnumItem` (`ayon_server/models/enum_item.py`, moved there from `ayon_server/enum/enum_item.py` to avoid a `forms`<->`enum` import cycle, re-exported from its old location) gained a `badges` field and can be passed directly as an option, coalesced down to the sparse shape.
- Companion frontend PR: https://github.com/ynput/ayon-frontend/pull/2311

## Test plan
- [x] `mypy`/`ruff` clean on `ayon_server/forms/`, `ayon_server/enum/`, `ayon_server/models/enum_item.py`
- [x] Manual smoke tests: rule/forward-ref/patch-key guards raise as expected; `normalize_options` round-trips strings/dicts/`EnumItem`s (incl. `IconModel` icons) into the sparse shape; all import orderings (forms-first, enum-first, via `api.enums` router) succeed without the former circular-import error
- [ ] Live end-to-end check against a running server (the `attrib` enum resolver needs a DB connection, not available in the sandbox this was built in)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NDaZ3GKnPbbjHoVWREtwCA


ayon_server/enum/base_resolver.py
```diff
@@ -21,7 +25,7 @@ async def get_accepted_params(self) -> dict[str, AttributeType]:
         """Return a dictionary of accepted parameters and their types."""
         return {}
 
-    async def get_settings_form(self) -> SimpleForm | None:
+    async def get_settings_form(self) -> "SimpleForm | None":
```
kalisp, line 28: Just out of curiosity, why the quotes? I found it is because compatibility with Python 3.9, but we are not planning to go back, are we?

martastain: Quotes are required when you import the type in `if TYPE_CHECKING` block. because then the object (that is used just for type hints is not actually imported during runtime.

ayon_server/forms/simple_form.py
```diff
@@ -48,6 +86,134 @@ class FormFileData(TypedDict):
...
+    if isinstance(value, dict):
+        result: set[str] = set()
+        for v in value.values():
+            result |= _template_vars(v)
+        return result
+    if isinstance(value, (list, tuple)):
+        result = set()
+        for v in value:
+            result |= _template_vars(v)
+        return result
```
kalisp:
```suggestion
    if isinstance(value, (dict, list, tuple)):
        items = value.values() if isinstance(value, dict) else value
        result: set[str] = set()
        for v in items:
            result |= _template_vars(v)
        return result
```

ayon_server/enum/base_resolver.py
kalisp: One needs to know or it will just blow up during runtime? I was suspecting it is because of TYPE_CHECKING, remove quotes, ran mypy and it wasn't complaining. (Which I was expecting it would after quotes removal.)

(But no biggie, just out of curiosity.)

martastain: It will blow up... TYPE_CHECKING is evaluated as False, during runtime, so it is never imported. It works with older pythons, because it treat type hints internally as strings.


https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports

kalisp (approved):
Tested with https://github.com/ynput/ayon-frontend/pull/2311

Worked fine.

PR #1117, opened 2026-09-11
Fix entity import permissions

martastain:
This pull request updates the permission checks in the `import_data` function to provide more granular control over who can import data, especially when a `project_name` is specified. The main change is that users who are not managers but have full hierarchy write access to the project can now use entity import

### DIFF
--- a/api/data_import/import_data.py
+++ b/api/data_import/import_data.py
@@ -170,11 +170,20 @@ async def import_data(
     Returns:
         ImportStatus: Summary of import results
     """
-    if not user.is_manager:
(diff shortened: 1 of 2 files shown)


kalisp:
Is there opposite front end PR to test it?

Now non admins cannot drag in file, without any log. Just with ` http://localhost:5000/api/csv/import/upload? 403 (Forbidden)` in developer tools.

martastain:
@kalisp fixed :crossed_fingers:

kalisp (approved):
With `Restrict Folder Create` off even artist could import data

Image

PR #1118, opened 2026-09-11
Resolving icon url for actions in enum

kalisp:
## Description of changes

Resolves `{addon_url}` in AYON based icons for actions with appropriate `addon.name` and `addon.version`.

### Additional context

Use `api/enum/actions` to check addon with actions with AYON based icons for actions. Most likely only `batchdelivery`.

image

### DIFF
--- a/ayon_server/enum/resolvers/enum_actions.py
+++ b/ayon_server/enum/resolvers/enum_actions.py
@@ -43,6 +43,11 @@ class ActionsEnumResolver(BaseEnumResolver):
                         if action.identifier in action_idents:
                             continue
 
+                        if action.icon and action.icon.url:
+                            action.icon.url = action.icon.url.format(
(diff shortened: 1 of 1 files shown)


martastain:
Shouldn't addons provide the urls themselves? They have all the necessary information.

filipvnencak (approved):
From FE POV working!

kalisp:
> Shouldn't addons provide the urls themselves? They have all the necessary information.

You mean that definition in addon should be already resolved? During `create_package`? That could be done, but it will be separate PRs to various repos. (Actually it seems that is only `batchdeliver` who is using this so far.)

But this could be used as a fallback for missed resolutions.

image

martastain:
I don't understand. for the relative path you only need addon name and addon version - both is set on the addon instance, no?

martastain (approved):

